PR #6, opened 2024-06-13
TensorFlow Implementation

maggiehollis:
Added in Dockerfile and tensor files to run TensorFlow


tools/tensorflow/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.11
+RUN apt-get update
+RUN apt-get install -y build-essential
+RUN apt-get install -y libhdf5-dev
```
samestep, line 4: In general you want to put `apt-get update` and `apt-get install` commands in the same `RUN`; take a look at this doc page for more explanation about why: https://docs.docker.com/build/building/best-practices/#apt-get

tools/tensorflow/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+RUN pip install h5py
+RUN pip install tensorflow-aarch64
```
samestep, line 6: Does this actually work for x86 too?

tools/tensorflow/tensor_functions.py
```diff
@@ -0,0 +1,3 @@
+def double(x):
+    y = x * x
+    return y
```
samestep, line 3: This function implementation isn't quite correct; the goal is that the function _itself_ should effectively return `2*x`, but that its implementation should do so by taking the gradient. In other words, the machinery for taking gradients should happen in here, not in `tensor_run.py`. You can take a look at how we do this in `tools/pytorch` or `tools/zygote` to compare.

tools/tensorflow/tensor_run.py
```diff
@@ -0,0 +1,43 @@
...
+    func = resolve(params["name"])
+    vals = [tensor(arg["value"]) for arg in params["arguments"]]
+    
+    
+    with tf.GradientTape() as tape:
+        tape.watch(vals)
+        ret = func(*vals)
+        
+    start = time.perf_counter_ns()
+    gradients = tape.gradient(ret,vals)
```
samestep, line 28: As mentioned in `tensor_functions.py`, this code shouldn't actually take the gradient; it should just call the function, and the implementation of the function it calls should be in charge of taking gradients.

samestep (changes requested):
Great work, thanks so much for debugging the TensorFlow installation! I had a comment about best practices in the Dockerfile, a question about cross-platform usage, and a correction about the semantics of the `double` function.

In addition, can you also create a new job in `.github/workflows/build.yml` which builds, uploads, and runs the TensorFlow Docker image, similar to what we already do for PyTorch and Zygote? That way, since the GitHub Actions runners are x86, we can also gain more confidence that this Docker image works on x86 machines too.

tools/tensorflow/tensor_functions.py
samestep: I just pushed commit 5f7aaba070a26cbe412bccbe39e197d2455c1f18 which adds a second function, `square`, that should clarify the intention here.

tools/tensorflow/Dockerfile
samestep: If it turns out that it doesn't, you can try just doing different behavior on each architecture by making use of the `TARGETARCH` build argument that Docker automatically provides: https://docs.docker.com/build/guide/multi-platform/#platform-build-arguments

tools/tensorflow/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.11
```
samestep, line 1: Interesting, so did you switch to this from `-slim` because it fixed the issues you were having?

samestep (approved):
Looks great, thank you!

tools/tensorflow/Dockerfile
maggiehollis: I tried slim early on but it seemed to cause a problem with tensorflow from the beginning, what fixed it was all the previous imports meant I no longer needed to import -aarch64 and could go with the standard that ended up being compatible with both.

samestep: OK! So just to clarify, are you saying that you need the non-slim version for it to work correctly? Or are you saying that it'd work now if you switch it back to slim?

maggiehollis: You need non-slim on mac but I think the slim version does work on x86. I could change the Dockerfile to download slim for x86 if you think it would be worthwhile.

samestep: Nope that's fine! If ARM needs non-slim then let's use that. Thanks!

PR #10, opened 2024-06-17
Clean up local and CI Docker builds

samestep:
With #6 we saw that building cross-platform Docker images can be slow, which is inconvenient for local development and testing. This PR makes `build.sh` single-platform and documents the change in `CONTRIBUTING.md` (also making a minor change to the Git clone instructions along the way). The multi-platform build is moved to `cross.sh`, which is now used by CI in a [matrix job](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) so that future tools are easier to add.


maggiehollis:
Looks good! So locally we can now run it on a single platform but actions still runs both?

PR #12, opened 2024-06-17
Generate CI tool list programmatically

samestep:
This PR adds a little Python script to automatically generate the list of `tools` for Docker builds in CI, so now each new tool only needs to touch its own subdirectory under `tools`. This will help reduce merge conflicts.

### DIFF
--- a/.github/matrix.py
+++ b/.github/matrix.py
@@ -6,7 +6,8 @@ from pathlib import Path
 
 def main():
     tools = sorted(p.name for p in Path("tools").iterdir())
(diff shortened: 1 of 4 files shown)


maggiehollis:
Nice! That'll definitely be helpful in the long run. What did you do that made the Actions so quick?

samestep:
@maggiehollis they're currently so quick because they're not working haha 😅 as you can see by which jobs are (not) running. Should be fixed in just a sec!

PR #13, opened 2024-06-18
Tapenade Implementation

maggiehollis:


samestep:
I assume this is still in-progress so I converted it to a draft; feel free to mark it as "ready for review" whenever it is ready.

Tip: I'd recommend using Python's [`subprocess`](https://docs.python.org/3/library/subprocess.html) module instead of [`os.system`](https://docs.python.org/3/library/os.html#os.system).

run.sh
samestep, line 1: Can you undo the diff you have on this file? I still think we should try to make it more convenient to do local development inside the Docker container, but that should probably go in a separate script instead of modifying `run.sh`.

samestep (changes requested):
This is great work! Thanks for figuring out how to run the build steps in your Python scripts, since this tool operates pretty differently from the ones we've incorporated so far.

Just one change we need before we merge this: currently, the `return` and `nanoseconds values are strings [(example)](https://github.com/gradbench/gradbench/actions/runs/9617995890/job/26530959776?pr=13):

```json
{"outputs": [{"return": "9.0", "nanoseconds": "90"}, {"return": "6.0", "nanoseconds": "90"}, {"return": "9.0", "nanoseconds": "90"}, {"return": "6.0", "nanoseconds": "90"}, {"return": "9.0", "nanoseconds": "81"}, {"return": "6.0", "nanoseconds": "80"}, {"return": "9.0", "nanoseconds": "90"}, {"return": "6.0", "nanoseconds": "90"}, {"return": "9.0", "nanoseconds": "90"}, {"return": "6.0", "nanoseconds": "100"}]}
```

Can you change it so that they are numbers instead, like we do for the other tools?

```json
{"outputs": [{"return": 9.0, "nanoseconds": 90}, {"return": 6.0, "nanoseconds": 90}, {"return": 9.0, "nanoseconds": 90}, {"return": 6.0, "nanoseconds": 90}, {"return": 9.0, "nanoseconds": 81}, {"return": 6.0, "nanoseconds": 80}, {"return": 9.0, "nanoseconds": 90}, {"return": 6.0, "nanoseconds": 90}, {"return": 9.0, "nanoseconds": 90}, {"return": 6.0, "nanoseconds": 100}]}
```

samestep (approved):
Thanks for the fix! Looks good to me!

PR #14, opened 2024-06-18
Write a basic parser

samestep:
This PR uses [Logos](https://github.com/maciejhirsz/logos) and [Chumsky](https://github.com/zesterer/chumsky) to make an initial parser for GradBench's source language which we're calling Adroit, with [Ariadne](https://github.com/zesterer/ariadne) for error reporting, then uses [Serde JSON](https://github.com/serde-rs/json) to insert the resulting AST into our `gradbench.json` config. The PR also adds a VS Code extension with syntax highlighting for Adroit.


avibryant:
Your .adroit file reads like Scala, :100:

samestep:
@avibryant yeah I pondered the syntax a few months ago until I found something I was happy with, and then when I looked around I realized that Scala already does most of the same things :)

maggiehollis (approved):
Everything works for me!

PR #17, opened 2024-06-21
Add SciLean

lecopivo:
This PR [installs Lean](https://github.com/leanprover/elan/tree/v3.1.1?tab=readme-ov-file#manual-installation) in a Docker image and then adds [SciLean](https://github.com/lecopivo/SciLean) as a tool. SciLean depends on [mathlib4](https://github.com/leanprover-community/mathlib4) which is apparently huge, so this is the largest Docker image we've built so far: I needed to add in a [GitHub Action that removes preinstalled software from the runner](https://github.com/AdityaGarg8/remove-unwanted-software/tree/v3) to get the build to work. Even then, running mathlib's `lake exe cache get` command for ARM using QEMU still fails in GitHub Actions due to some network error, so this PR also adds a new `docker` folder to hold base images like that which need to be built manually, and then has the SciLean `Dockerfile` pick up after the mathlib4 build cache has already been populated. Even taking advantage of that, the SciLean Docker image build itself still takes over two hours.


maggiehollis:
Works on my computer! Only thing is the formatting of the output makes a new line for each return value and time pair. I don't know if this is relevant but just something I noticed since it's a bit different.

samestep:
@maggiehollis sweet, thanks for looking! Yeah I think it should be fine; if you have [jq](https://jqlang.github.io/jq/) installed, you can try piping the output to it:

```sh
./run.sh scilean | jq
```

This would have jq reformat the output so it'd look more standard. You can do the same with our other Docker images too.

PR #19, opened 2024-06-24
Implement DiffSharp

maggiehollis:


tools/diffsharp/run.fsx
```diff
@@ -0,0 +1,42 @@
...
+let run (pars : JsonValue) =
+    let inputs = pars?arguments
+    let values = [for item in inputs -> dsharp.tensor (item?value.AsFloat())]
+    let name = pars?name.AsString()
+    let stopwatch = Stopwatch.StartNew()
+    let result = 
+        if name = "square" then square (values.Head)
+        else double (values.Head)
+    stopwatch.Stop()
+    (float result, float stopwatch.ElapsedTicks)
```
samestep, line 22: Couple questions:

1. Could you point me to where in the documentation it clarifies that this means nanoseconds?
2. Could you change this so that the stopwatch starts after branching on `name` and getting `values.Head`, so it only times running the function itself and not resolving the function name?

samestep (changes requested):
Awesome job, thanks @maggiehollis! Just had one question and one request.

tools/diffsharp/run.fsx
maggiehollis: Just pushed the updated timing! And it's not directly in the documentation but I can show you how I figured it out in the office

samestep (approved):

PR #20, opened 2024-06-24
SciLean update: run AD benchmark in IO

lecopivo:
You can now run `benchAD` function but calling `runBenchAD` in `main` function.

### DIFF
--- a/tools/scilean/Main.lean
+++ b/tools/scilean/Main.lean
@@ -1,5 +1,6 @@
 import Lean
 import SciLean
+import Mathlib.Lean.CoreM
 
 open Lean
 open Except FromJson ToJson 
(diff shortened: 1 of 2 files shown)


samestep (approved):
Awesome, thanks Tomáš! Is `supportInterpreter` documented anywhere?

lecopivo:
Here is the [doc](https://github.com/leanprover/lean4/blob/875e4b1904f9eb41630100dd22d689c3cc65a0ae/src/lake/Lake/Config/LeanExeConfig.lean#L60), btw you can open up lakefile in VS Code and hit "go to definition" on `supportInterpreter` and it will take you there. You can also just hover over `supportInterpreter` with your mouse and a pop up window with doc shows up.

PR #23, opened 2024-06-26
ADBench GMM for PyTorch

maggiehollis:
This PR implements the GMM equation from [ADBench](https://github.com/microsoft/ADBench) for PyTorch. Additionally, it adds in error handling so tools do not crash if they don't support a given module (such as GMM).


tools/diffsharp/modules.fsx
samestep, line 1: Instead of having `runModule` throw an error using `failwith` and then having a separate `moduleExists` function to check whether it exists, can you change it to just be one function that returns an [`option`](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/options)? Then you can use [`Option.isSome`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html#isSome) to check whether the module exists for `"define"` messages, and [`Option.get`](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html#get) to get the value (probably the function itself; see my message on `run.fsx`) for `"evaluate"` messages.

samestep (changes requested):
Thanks for your great work on this @maggiehollis! Great to have the new GMM eval in our reusable JSON format, and great to now have the ability for tools to respond with their success or failure when they load a module.

I have a few comments and requests; don't be too alarmed, this is just because this PR is larger-scale than previous ones you've done. I also realize that I haven't done the best job at communicating all of these over written text, so when you get to this on Monday or Tuesday, we can definitely meet to talk about any of these if you'd like.

Have a great rest of your holiday weekend!

samestep:
Ah right, the one thing I forgot to mention: nice job making the GMM eval work correctly for SciLean since it doesn't return a `"success"` field! As we previously discussed, I'm happy to modify our Lean code so it returns that field to be consistent with the other tools. Would you prefer for me to do that on this branch before we merge this PR, in a separate PR after we merge this one?

samestep:
Oh also, could you rename the title of this PR to clarify that it's only GMM, not all the ADBench equations?

tools/pytorch/utils.py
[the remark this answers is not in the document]
samestep: @maggiehollis thanks for pasting in the license text! One more request: for each of the files you copied in, can you also add a link to the file on GitHub that you copied it from? That way it's easy to directly see where everything came from.

tools/diffsharp/modules.fsx
samestep, line 1: Thanks, this is close! Still not quite what we want, though. For comparison, take a look at the `resolve` function I wrote for Lean:

https://github.com/gradbench/gradbench/blob/227073148831f0082b33c2d9aa6b289c818483f1/tools/scilean/Main.lean#L116-L123

Rather than taking in both the module name and function name at once, this Lean implementation takes in the module name first, then returns an `Option` of a function which takes in the function name, which also returns an `Option` of the function itself. Can you do the same thing here?

The goal is to avoid having two separate sources of truth for resolving just the module versus resolving the module followed by the function.

samestep (changes requested):
Thanks for quickly addressing all my earlier comments! I just had two more requests, one here about F# and one about GitHub links in response to my earlier comment about license text.

Also, when you have the chance, could you respond to the remaining unresolved comments I left earlier? Not a blocker for this PR, but good to get the ball rolling.

samestep (approved):
Looks awesome! Let's merge 😃

PR #41, opened 2024-07-18
Initial BA implementation for PyTorch

maggiehollis:
Implements BA equation from [ADBench](https://github.com/microsoft/ADBench/tree/38cb7931303a830c3700ca36ba9520868327ac87).

### Memory:
- ADBench ran their benchmarking on a machine with 32GB of RAM
- We are running on a machine with 16GB of RAM
- Memory should not currently be an issue as ADBench repo (with default settings) for BA only used up to around 14GB
  - Default settings are only running on files ba1 through ba5
  - Default repo has measures in place to stop running a function if it uses up too much memory but they currently have the max set to infinity
  - Unclear what (if any) memory limitations were used for official benchmarks
- On our machine, the jacobian calculation for ba2 took 3GB, ba3 took 11GB before it timed out (after 300 seconds)

### Timing:
- ADBench has two different timing messages
  - Killed after 300 seconds
    - Kills process in run-all.ps1 (line 200) if call is taking too long
    - Sets status to timed out so future runs don’t happen (see below)
  - Didn’t run due to guaranteed timeout
    - If previous runs timed out these will automatically be skipped
- The ADBench repo with its default settings has a 300 second timeout in place around individual functions
  - ba1 and ba2 run, ba3 times out, ba4 and ba5 are then skipped
  - I could not find what timeout they used for their published benchmark
- I implemented a 300 second timeout
  - Set up a signal alarm that goes off if function hasn’t completed in 300 seconds
  - Jacobian for ba3

### Segmentation Fault:
- Despite the timeout and memory staying within ADBench's limits, some files throw a segfault
  - Objective for ba3: successfully runs in around 100 seconds but throws a segfault before output is sent over JSON
  - Other files throwing a segfault as well right when they reach the timeout, this is likely due to how the process is killed using signal. Using multiprocessing instead could fix this but causes other issues.

### Other:
- ADBench published results include a couple tools we have implemented (DiffSharp, Autograd, etc.) but PyTorch is not one of them


evals/gmm/README.md
samestep, line 1: Looks like this was accidentally edited; could you revert your change to this file?

evals/ba/README.md
samestep, line 1: Thanks for the detailed writeup! One request: since the data files are copied directly from ADBench, could you create a `evals/ba/data/LICENSE` file copied from their `LICENSE` file, just to be extra clear?

evals/ba/main.py
samestep, line 1: I see that all the BA input files are super short, but currently the `input` data that we send across as JSON in the protocol is very long; instead of expanding out into multiple megabytes of JSON data that we send over text, could we change it so that it just parses the text file into something structured, then sends across that much smaller amount of data? Then on the other side, the tool can take in the smaller description and duplicate the points as necessary.

tools/pytorch/ba.py
samestep, line 1: Could you copy the ADBench license into the top of this file and also add a URL, similar to what you've already done for `tools/pytorch/gmm.py` etc?

evals/ba/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Generation
+
+Files are provided with information to generate the below inputs. 3 values, N, M, and P are used alongside one camera's paremters $c \in \mathbb{R}^{11}$, one point $x \in \mathbb{R}^{3}$, one weight $w \in \mathbb{R}$, and one feature $f \in \mathbb{R}^{2}$. N is the number of cameras, M is the numer of points, and P is the number of observations.
+
+### Inputs
+
+The data generator returns a dictionary with the following inputs
+
+1. Cams: Set of camera parameteres used. The parameters are rotation, camera center, focal length, principal point, and radial distortion. $c$ is duplicated N times to create
```
samestep, line 15:
```suggestion
1. Cams: Set of camera parameters used. The parameters are rotation, camera center, focal length, principal point, and radial distortion. $c$ is duplicated N times to create
```

evals/ba/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. Reproj-Error: Reprojection error of how well the projected 3D structure aligns with the given 2D image.
+
+$$reproj_{error} \in \mathbb{R}^{2P}$$
+
+2. W-Error: Regularization error ensures the weights are confined to a specific range. In this scenario it penalized $w$ if it deviates from 1.
+
+$$w_{error} \in \mathbb{R}^{P}$$
+
+3. Jacobian ($J$): How the reprojetion errir will change given adjustements to the above inputs.
```
samestep, line 45:
```suggestion
3. Jacobian ($J$): How the reprojection error will change given adjustments to the above inputs.
```

tools/pytorch/run.py
```diff
@@ -20,9 +21,19 @@ def tensor(x):
...
-        return ret.item()
-    return ret.tolist()
+    try:
+        if type(ret) == BASparseMat:
+            return str(f"BASparseMat: {ret.nrows} rows by {ret.ncols} columns")
+        if type(ret) == tuple:
+            reproj, w = ret
+            len_r = len(reproj.tolist())
+            len_w = len(w.tolist())
+            return f"[{len_r/w}{reproj.tolist()[:2]},{len_w}{w.tolist()[0]}]"
```
samestep, line 31: Instead of converting these to strings, can you make them into structured data? Similar to how the GMM `input` is a JSON object with multiple fields containing different parts of the data.

It'd probably make sense to do it similar to how we're doing it in DiffSharp, where each function (like `calculate_objectiveBA` and `calculate_jacobianBA`) can have its own functions attached, to convert the input from JSON and the output to JSON.

samestep (changes requested):
Thanks for doing the hard work to implement debug this! I look forward to seeing the writeup you mentioned of your notes describing the different things you tried and any differences from the original implementation in ADBench. I left a few comments and requests here for when you have the chance to address them.

tools/pytorch/run.py
maggiehollis: Sounds good! Do you mean something like `{"BASparseMat": {"rows": mat.nrows, "columns": mat.ncols }}` rather than the string? And `{"reproj_error": r_err.tolist(), "w_err": w_err.tolist()}`? For reproj and w error, they are huge lists where the first element (or two) are just repeated a tone of times. Do you still want me to do something like that to make it viewable or leave the whole list?

samestep: Yep that sounds perfect! Since you've mentioned that the error vectors are just the same numbers duplicated many times, you could just output the data point and also how many times it's duplicated. And for the Jacobian matrix, for now it's fine to just output the dimensions; maybe later we'll decide to change it so we serialize the sparse matrix data, but no need to do that right now.

samestep:
> Unclear what (if any) memory limitations were used for official benchmarks

Not a direct answer to this question but by [using `git blame`](https://github.com/microsoft/ADBench/blame/38cb7931303a830c3700ca36ba9520868327ac87/ADBench/run-all.ps1#L56-L57) we can find the PR where they introduced the OOM machinery: microsoft/ADBench#177

samestep (approved):
Awesome work. Let's merge it!

PR #45, opened 2024-07-24
Website Mobile View

maggiehollis:


packages/gradbench/src/App.tsx
```diff
@@ -3,38 +3,38 @@ import "./App.css";
...
-        allows for the easy addition of new functions and tools.
+        <li> GradBench is designed to work across languages and domains. </li>
+        <li>
+          Functions are written in <b style={{ color: "#3e4756" }}>Adroit</b>{" "}
+          and automatically translated to a tool's native language.
+        </li>
+        <li>
+          The architecture allows for the easy addition of new functions and
+          tools.
+        </li>
```
samestep, line 37: Having these as bullets is OK, but putting the `<li>` nodes inside of a `<p>` instead of a `<ul>` doesn't work; the indentation is broken when the `<li>` is split across multiple lines, like on mobile. I'm guessing that the reason you don't use `<ul>` is because this part of your CSS causes the bullets to disappear:
```css
* {
  margin: 0;
}
```
Can you modify your CSS so that it works? I'm guessing that you'll just need to replace the `*` with something more specific, or you could apply your own styling to `<ul>` or `<li>` to make it work.

samestep (approved):
Great work, thanks for doing this so fast! I had one request about formatting bullet points; otherwise I think this is good to go :+1:

packages/gradbench/src/App.tsx
```diff
@@ -50,138 +98,157 @@ const App = () => {
...
+            <b style={{ color: "#3e4756" }}>Hello</b> module that computes the
+            derivative of x&sup2;. <b style={{ color: "#3e4756" }}>Hello</b> can
+            be run on any of the tools listed in the table.
+            <br />
+            <br />
+            We are currently working on implementing functions from Microsoft's
+            ADBench suite. The Gaussian Mixture Model Fitting (
+            <b style={{ color: "#3e4756" }}>GMM</b>) equation is currently only
+            supported by PyTorch. Additionally, PyTorch's support for the Bundle
+            Adjustement (<b style={{ color: "#3e4756" }}>BA</b>) equation is
```
samestep, line 249:
```suggestion
            Adjustment (<b style={{ color: "#3e4756" }}>BA</b>) equation is
```

samestep (approved):
Great work! One tiny change and let's merge it.

PR #52, opened 2024-08-01
Decorator for inputs and outputs

maggiehollis:


samestep (approved):
Looks great! Nice work, thanks for doing it so fast :)

PR #54, opened 2024-08-02
Reinstall numpy for GMM

maggiehollis:
Accidentally deleted numpy for GMM eval but it was necessary for `data_gen.py`


samestep:
Aha, this was bound to happen sooner or later. The issue is that the `run` job in CI only starts once all the `eval` and `tool` builds are done, and while some tool builds are fast, the longest one (SciLean) takes over two hours. So we've gotten into the habit of merging PRs before we get any signal at all from `run`.

I think a potential solution to this would be to split up the `tool` job matrix into a "fast" portion and a "slow" portion, and then similarly split up the `run` job, so that we can get quick signal from all the evals with at least some of the tools, and then eventually get signal from the rest of the tools once it runs to completion.

Thoughts @maggiehollis?

maggiehollis:
Yeah that sounds like a good idea! Should I go ahead a merge it or do you want to make the actions changes in this PR?

output.txt
samestep, line 1: Can you delete this file?

samestep (changes requested):

samestep:
> Yeah that sounds like a good idea! Should I go ahead a merge it or do you want to make the actions changes in this PR?

@maggiehollis if you're confident that this PR is correct then we can just merge it!

Do you want to try making those changes to GitHub Actions, or would you like me to do it?

output.txt
maggiehollis: Yep! Sorry about that

samestep (approved):

maggiehollis:
Could you do it? I've never worked with actions before.

PR #55, opened 2024-08-02
Adding Wrapper to Python Tools

maggiehollis:
Updating Autograd, Jax, MyGrad, and TensorFlow with the wrapper function originally written for PyTorch. This wrapper is a [decorator](https://stackoverflow.com/questions/739654/how-do-i-make-function-decorators-and-chain-them-together) that has similar functionality to those for SciLean and DiffSharp.

Autograd inputs and outputs do not really require adjustment. Currently I have a lambda function multiplying by 1.0 to ensure floats. This is not strictly necessary.

Tapenade is written in a combination of Python and C but inputs don't need to be changed. Inputs are simply converted to strings (and multiplied by 1.0 to ensure they are floats) when sent to C files and converted back to floats to be returned in JSON. A wrapper could be implemented here but it is not needed.

Similarly Zygote appears to not need a wrapper.

### DIFF
--- a/tools/autograd/run.py
+++ b/tools/autograd/run.py
@@ -11,7 +11,7 @@ def resolve(module, name):
 
 def run(params):
     func = resolve(params["module"], params["name"])
-    arg = params["input"] * 1.0
+    arg = params["input"]
     start = time.perf_counter_ns()
     ret = func(arg)
     end = time.perf_counter_ns()
--- a/tools/autograd/run.py
+++ b/tools/autograd/run.py
@@ -11,11 +11,11 @@ def resolve(module, name):
 
 def run(params):
     func = resolve(params["module"], params["name"])
-    arg = params["input"]
+    input = func.prepare(params["input"])
     start = time.perf_counter_ns()
-    ret = func(arg)
+    ret = func(input)
     end = time.perf_counter_ns()
-    return {"output": ret, "nanoseconds": {"evaluate": end - start}}
+    return {"output": func.unwrap(ret), "nanoseconds": {"evaluate": end - start}}
 
 
 def main():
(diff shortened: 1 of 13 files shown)


maggiehollis:
@samestep This could be ready to merge but I want your opinion on using the wrapper for Autograd. I think we should keep it just to ensure all inputs are floats but it isn't super necessary. Also let me know what you think about a wrapper for Tapenade. I think we wouldn't really need one but it's up to you. I would probably put it in `gradbench.py` just to multiply vals by 1.0 and convert to a string.

tools/autograd/run.py
```diff
@@ -11,7 +11,7 @@ def resolve(module, name):
 
 def run(params):
     func = resolve(params["module"], params["name"])
-    arg = params["input"] * 1.0
+    arg = params["input"]
```
samestep, line 14: I notice that you don't call `prepare` or `unwrap` here; is that a mistake?

samestep (commented):
Looks great, thanks! Just had one question.

> @samestep This could be ready to merge but I want your opinion on using the wrapper for Autograd. I think we should keep it just to ensure all inputs are floats but it isn't super necessary. Also let me know what you think about a wrapper for Tapenade. I think we wouldn't really need one but it might be good to have, it's up to you.

@maggiehollis I think it's fine to leave out the wrapper stuff for tools that don't need it, but it also doesn't hurt to include it.

> I would probably put it in gradbench.py just to multiply vals by 1.0 and convert to a string.

What do you mean about converting to a string? Is this some sort of thing with the way we pass inputs to C programs in the case of Tapenade?

samestep (approved):

PR #71, opened 2024-08-29
Try using `windows-2019` instead of `windows-2022`

samestep:
Followup on #69, trying to figure out why the Windows VS Code extension built in GitHub Actions just gives `_` on hover but otherwise seems to work correctly; in contrast, when I build locally, everything works perfectly.


samestep:
Didn't work.

PR #87, opened 2024-09-26
Translator for Hello Eval

maggiehollis:
Working to translate a `hello.json` (generated from `hello.adroit`) file into a python file `hello.py` for PyTorch. Currently only have an initial idea of how to set up the process and how to go about the translation. 

One thing I'm still thinking about: How/where should the `hello.json` be generated?

Currently you need to run the below in the terminal. Then build the eval and tool and run as usual.

```sh
target/release/adroit json evals/hello/gradbench.adroit > hello.json
```


maggiehollis:
Since hello.json doesn't exist unless I make it, this won't run on actions.

samestep:
@maggiehollis to generate `hello.json` I think we can just put the `adroit` binary in the Docker image [like I described on Discord](https://discord.com/channels/1242864568850579487/1267596255886970952/1288199706048204831); does that work?

samestep:
You can also take a look at what I currently have in #85; looks like you and I have worked on different parts of this so far which is nice, shouldn't be too hard to combine what we've both already done. Specifically, these two lines take the Adroit source code from the eval and write it to a file:

https://github.com/gradbench/gradbench/blob/63c2746c28e930eedbd5d89d7d74096145e841a8/python/gradbench/pytorch/run.py#L30-L31

And then this line assumes `adroit` is on the `PATH` (as achieved by that Dockerfile `ADD` command I linked above) and runs it on the file we just wrote, to get the JSON:

https://github.com/gradbench/gradbench/blob/63c2746c28e930eedbd5d89d7d74096145e841a8/python/gradbench/pytorch/run.py#L35

maggiehollis:
Finished setting up code so `hello.json` would be automatically generated. Additionally set up a general structure for the code (check if module exists, if not try to translate, otherwise return false). I temporarily renamed `hello.py` to `hello_temp_rename.py` so I could skip straight to translation.

python/gradbench/pytorch/run.py
```diff
@@ -18,6 +20,41 @@ def run(params):
...
+        raise ValueError(f"resolved path {path} is not relative to {folder}")
+
+
+def define(module, source):
+    try:  # File already exists
+        import_module(module)
+    except:  # Try to translate to create file
+        src = adroit_path(module)
+        src.write_text(source)
+        adroit = "usr/local/bin/adroit"
```
samestep, line 40: Nice work on this!

One request: currently this only works in the Docker image because the cwd is set to `/`, but when running locally via Poetry, it breaks because this path isn't absolute. You actually don't need to specify the full path at all though, because the `ADD` command in the Dockerfile puts it on the `PATH`, so instead of specifying the full path you can just say `adroit`:

```suggestion
        adroit = "adroit"
```

samestep:
See #123.

PR #90, opened 2024-10-22
Tapenade Compilation Fix

maggiehollis:
Ensures Tapenade only compiles once during the `define` message and then runs that compiled code for `evaluate` calls.


samestep (approved):
Seems good to me! Un-draft and merge whenever you're ready :)

PR #91, opened 2024-10-22
Futhark implementations of benchmarks

athas:
This is WIP. No Docker image and such is built yet. I'll also implement the other benchmarks. The Python `run.py` should probably also have some of its duplication reduced.


athas:
This now has implementations of all ADBench benchmarks (although plumbing for HT is missing, because it's not in GradBench yet). They all work, but don't return the results in the same format as the PyTorch versions yet, because it's not documented what they should look like, but eyeballing I can see that the same values are present.  The implementations produce the exact values (in the exact format) that the original ADBench tooling is looking for.

athas:
This now contains Futhark implementations of BA, GMM, and hello. A Docker image is also included. Unfortunately it is currently x86-64-specific, and making it also support ARM would significantly increase the Dockerfile building time (because the Futhark compiler would have to be compiled from scratch).

samestep (approved):
This is awesome, thanks so much for the contribution! Apologies for the missing HT and lack of documentation; as discussed on Discord, we'll work on adding those soon.

Issue #98, opened 2024-10-22
Validate results from each evaluation


samestep:
Currently, CI just builds the evals and tools and then runs all pairs. There is no signal about whether a tool succeeded or failed on a given eval. We want to designate a specific tool as the "golden" reference implementation (the plan is to use PyTorch), and then for each eval, define a way to compare the results of two tools on that eval. This also relates to #97, since we'll want to visualize those validation outputs on the website. Beyond this, we probably also want each eval to define how two runs should be visualized in more detail, beyond just a binary "correct/incorrect" decision.

athas:
Validation needs to be part of an eval, since as far I understand it, `run.py` is just a very generic message broker, that doesn't actually understand anything about the messages it transmits.

athas:
On the other hand, all an eval can do is send messages to the tool, so how is it supposed to report validation errors?

samestep:
@athas `run.py` can (and I think probably should) validate the protocol itself, but the eval needs to describe how to validate the data it got from the session.

One way to do this could be to add an "end" message to the protocol, so when the eval wants to terminate the session (either successfully or unsuccessfully), it can send a message that `run.py` doesn't forward to the tool and instead just logs and exits. The eval could accumulate whatever it wants from the session into rich metadata in this final message, and the protocol could define some standard fields to be visualized consistently on the website (e.g. success vs failure, overall performance, etc).

athas:
I have also come to the conclusion that making `run.py` somewhat protocol-aware is necessary. I think this is fine as GradBench is not supposed to be a completely generic testing tool. The most important thing is to make it easy to add new tools, and achievable to add new evals.

I don't think having an "end" message is a problem, but I don't think it's how to do validation. Now I'm considering a design where `evaluate` messages must have two fields that are known to `run.py`:

1. `name`, which is a human-readable string unique to the particular input.
2. `input`, which contains the actual input to the evaluation (currently a JSON object, may become a filename as per #99).

The response must then contain another field known to `run.py`:

1. `output`, which contains the result.

(You will not that `input`/`output` are actually the fields already used by GradBench.)

[...]

PR #100, opened 2024-10-22
Choose PyTorch as a "gold" tool to run first

samestep:
As a first step for #98, this PR adds a `gold` job to our Build and Nightly workflows, which runs a single tool on all the evals first before running any of the other tools, so that tool's results will be available to validate all the others' against it. In `.github/matrix.py` we specify that our particular choice for this "gold" tool is PyTorch; quoting @athas [on Discord](https://discord.com/channels/1242864568850579487/1267596255886970952/1298393783033139402):

> I would probably use PyTorch as the golden implementations, because it's so popular, so there's a pretty good chance people will be familiar with it. Could also be autograd.

Note that currently, this CI setup runs PyTorch _twice_ on every eval, since it's also a part of the "tool" and "fast" sets that get run after the "gold" tool. This is intentional: even though it's a bit redundant, it'll serve as a nice easy determinism smoke check, since while for other tools there might be some numerical differences, we'd expect that running PyTorch twice should be exactly the same.

### DIFF
--- a/.github/matrix.py
+++ b/.github/matrix.py
@@ -22,6 +22,7 @@ def main():
     slow = ["scilean"]
     output("fast", sorted(tool - set(slow)))
     output("slow", slow)
+    output("gold", "pytorch")
 
 
 if __name__ == "__main__":
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,7 @@ jobs:
       eval: ${{ steps.matrix.outputs.eval }}
       fast: ${{ steps.matrix.outputs.fast }}
       slow: ${{ steps.matrix.outputs.slow }}
+      gold: ${{ steps.matrix.outputs.gold }}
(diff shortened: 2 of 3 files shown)


athas:
For the record, the downside of PyTorch as a golden implementation is that it is a complex dependency. It is possible that an update of PyTorch will at some point in the future cause all *other* tools to fail. I still think the tradeoff is worth it, however.

samestep:
@athas agreed. Another consideration is that, after chatting with @martinjm97 this week, I think it'd be good to be able to support programs that include integrals, since those of course come up in differentiable rendering and other applications. PyTorch won't really work for this. So maybe what we want is for the "golden" tool to be separate, and let it choose PyTorch when it makes sense but allow to choose others.

Currently this is very easy to swap out by just editing `.github/matrix.py`; hopefully we can keep it that way.

Issue #99, opened 2024-10-22
Support file sharing between evaluations and tools


samestep:
Currently the only way to pass data between an eval and a tool is via JSON through the protocol. This becomes a bit unwieldy for evals that need to send a lot of input data or that require the tool to send back a lot of output data. Perhaps we should also allow a session to include a shared directory that the eval and tool can both and read and write.

This raises a couple questions:

1. How can the protocol (currently encapsulated in `run.py`) expose this in a way that remains agnostic of the way the eval and tool are implemented? Currently the eval and tool just need to be commands that can be run to produce processes with stdin and stdout, but this change would require them to have more complex interactions with the outside world. It might be fine though; we already have `eval.sh` and `tool.sh` pass the `--interactive` flag to `docker run`, so maybe we can just do a similar thing for a filesystem mount.
2. Should filesystem events be reflected in the protocol at all? Or should it simply be conventional? For instance, perhaps the eval can make sure to write an input file before it sends an `"evaluate"` command, and require that the tool write its output file before it sends its response.
3. How should output files be saved for later analysis and validation? Currently the entire session can be saved as one JSON file, but with this change, the filesystem could potentially undergo destructive writes in the course of a session. Should we just save all the files in the final state of the shared directory? Or should we impose some limitations that ensure that, e.g., files can only be written once and not overwritten?

athas:
I have an idea for how to do this. If the protocol is modified such that the input and output data is represented as file names in the messages, then the `tool.sh` automation (or something it invokes) can automatically detect when a Docker image is invoked, and transparently copy files and and forth and rewrite the file names in the messages. This means the automation needs to be somewhat protocol-aware rather than just a dumb proxy, but the important thing is that specific tool implementations don't have to care - it will all be completely transparent.

I think persistent storage of output (and input) files should be done on the "host", understood as the system that runs `run.py`, outside the Docker images. Actual validation can be done in Docker images, by sending files there, using the same idea of transparently copying files and rewriting file names.

samestep:
@athas this roughly lines up with what I was thinking, and I like the idea of handling it in `tool.sh`. How difficult would it be to throw together a prototype? If it's not too hard, I think I'd have an easier time deciding whether this makes sense upon looking at an initial implementation of the idea.

athas:
I'm not much of a Docker expert, but I'll give it a shot.

samestep:
[...]

PR #101, opened 2024-10-26
Validation of results

athas:
Now run.py will dump inputs and outputs for every evaluation. All evaluation message must be extended with a 'workload' field, and all tools must respond with a 'tool'.

There is no actual validation yet.


athas:
I now have a validation script as well. The intended workflow is that you first run the tools you care about, e.g.:

```
$ /run.py --eval './eval.sh hello' --tool './tool.sh futhark'
$ /run.py --eval './eval.sh hello' --tool './tool.sh pytorch
```

This will populare a directory called `results` (can be changed with command line option).

After execution you run validation for the specific eval, where you specify which tool is golden and which other tool you want to check:

```
$ evals/hello/validate.py  --eval hello --golden pytorch --tool futhark
```

Since everything is file based and decoupled, it's easy to copy things around as needed. It's also easy to imagine a run-everything script (like for CI) that will automatically do the validation.

athas:
Validation now works for GMM, BA, and Hello. The Futhark implementation validates against the PyTorch implementation. I had to make some other changes (fix some spellings, ensure deterministic GMM input data) to make that work. I think it is time to discuss whether this approach is the right one.

samestep:
Looking at this now; sorry for the delay! I think it'll take me a bit to page back in all the context and think about the design more carefully.

samestep:
Marking this as draft for now since CI is failing, but don't worry about fixing it at the moment (I'm still working on some other comments and suggestions).

samestep:
OK @athas what do you think of the approach in e0fbea1c229d6d39fec35cd6f6bb52e8478e625b? Specifically, I'm not a huge fan of the added complexity from the `"workload"` and the `"results"` file structure, since one reason I had `run.py` echo the JSON log to stdout is so that the log contains all the information that was exchanged between the eval and the tool; that's why in CI we pipe it to `tee`.

So, in this alternate approach, we instead just feed the validation script two JSON logs: one from the golden tool, and one from the tool being validated. Then, in `python/gradbench/validate.py`, I've written a bunch of types that can be used with [Pydantic](https://docs.pydantic.dev/latest/) to validate the protocol itself, as well as a pretty general `validate_fixed` function that can be reused across all evals that are simple like the ones we currently have: they essentially just send a fixed sequence of messages in the same order. So then, all three of our current validators have basically the same `main` function, which just parses two CLI args and calls `validate_fixed` with a custom `checker` that matches on the `name` of the function and uses that to select which method to check the outputs of the two tools against each other.

Thoughts?

samestep (changes requested):
@athas separate from the design, it looks like [the BA validation is failing](https://github.com/gradbench/gradbench/actions/runs/11644274561/job/32426161112?pr=101); when you have a chance, can you take a look?

> TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

athas:
How will validation based on the protocol log work with #99?

athas:
I like the conceptual simplicity of verifying the log, despite it being somewhat unusual. I can think of the following challenges:

* Results that are external to the log for size reasons, such as above.
* Verifying partial results - e.g. maybe you are running only a subset of the workloads for some reason, but still want to validate those.

And one potential benefit:

* Low-latency validation, simply by validating the log messages immediately as they come in.

The above are not immediate showstoppers and have technical solutions. I suggest that you merge the log validation approach (once I've looked at BA) and then we deal with the rest as it becomes relevant.

samestep:
@athas

> - Verifying partial results - e.g. maybe you are running only a subset of the workloads for some reason, but still want to validate those.

Can you clarify what you mean by this? For instance, what might an example of that look like? Using this design, an eval and its validator can still work however they wish, so if the eval somehow decides to only run a subset of things, its validator can take that into account. The `validate_fixed` helper function we have currently is just to simplify what I expect to be the "common case".

> - Low-latency validation, simply by validating the log messages immediately as they come in.

Interesting idea. To add more context, let me first describe a different idea and then explain how it relates. Currently, the validators all depend on some shared code and the Pydantic library, so e.g. the `run-fast` and `run-slow` CI jobs need to install Poetry to install Pydantic, then just assume every eval has a `validate.py` script in its folder in this repo, and just run that. This was fine for prototyping, but it goes against our otherwise-modular architecture of putting everything in its own Docker image.

So, instead, I'd like to put the validator inside of the Docker image for its eval. This should work fine since `docker run` lets you specify an `--entrypoint`. But to pass the two log files in, we'd need to also run the Docker image with some sort of bind mount, and then have a convention for where the log files should be.

That's not a bad solution, but it did get me thinking back to what I wrote in https://github.com/gradbench/gradbench/issues/98#issuecomment-2439596813:

> One way to do this could be to add an "end" message to the protocol, so when the eval wants to terminate the session (either successfully or unsuccessfully), it can send a message that `run.py` doesn't forward to the tool and instead just logs and exits. The eval could accumulate whatever it wants from the session into rich metadata in this final message, and the protocol could define some standard fields to be visualized consistently on the website (e.g. success vs failure, overall performance, etc).

A modification of that idea would be that `run.py` itself takes an optional CLI argument called `--golden` that can point to a file containing the log from the golden tool. Then at the end of a session, the eval sends a message saying `"kind": "end"` or something, at which point `run.py` shuts off the tool. If no `--golden` was specified then `run.py` response to this `"end"` message by just telling the eval not to do any validation. But if `--golden` was specified, `run.py` can just send the entire golden log (since it's just JSON) to the eval, and then the eval can validate the session it just ran against that golden log, and return its result as structured data (probably in a last message labeled `"kind": "analysis"` or something), which will then become part of the final log before exiting.

This has a couple benefits:

- Everything stays within the protocol, only using JSON over stdin and stdout for communication.
- Instead of only having access to two logs, the eval could potentially keep track of other internal state, and possibly use that to make validation easier?

Alternatively, instead of sending an `"end"` message, the eval could send a `"kind": "start"` message at the beginning, and `run.py` could send the golden log _then_ (if it has one). If I understand correctly, this would enable what you suggested of validating responses immediately as they come in.

It does have a downside, though: because the eval has access to the golden log at the beginning, this means that it could potentially use the golden log to affect what it does in the session. I can't think of why someone would do that, so maybe it's not a big deal, but it feels like it'd just be simpler for the validator to only see the golden log at the end of the session.

In any case, for now I think I'm going to see how quickly I can implement the other idea of just having the validator be a separate entrypoint script. Actually, I don't think I even need to do the bind mount thing: I could just make some top-level `validate.py` script on the host that sends the two JSON logs directly to the eval's validator via stdin.

Although, in response to your other point:

> - Results that are external to the log for size reasons, such as above.

To handle those sorts of external files, I guess a bind mount would be necessary anyway. But as you suggested, I'd prefer to first merge an initial implementation of this that just uses the singular JSON log file without any external files, and then revisit the design if #99 turns out to be really necessary.

athas:
> Can you clarify what you mean by this? For instance, what might an example of that look like?

I might download a golden log file of GMM evaluating all its one thousand (or whatever) workloads, some of which are very large and memory consuming. But for my own new tool, I only want to check that ten representative ones work, because I can't run all the workloads (or don't want to wait for them to finish). This is pretty common use of ADBench. It can be implemented by performing some surgery on the log file (or by implementing your suggestion of making the eval itself perform the validation relative to a provided log file, which could be used in a dictionary-like fashion).

> It does have a downside, though: because the eval has access to the golden log at the beginning, this means that it could potentially use the golden log to affect what it does in the session. I can't think of why someone would do that, so maybe it's not a big deal, but it feels like it'd just be simpler for the validator to only see the golden log at the end of the session.

This is not a big deal. Evals are trusted. They must be, as they are ultimately responsible for validation as well.

athas:
I fixed the validation of BA, but I have no idea what to do about the scilean timeout.

athas:
Everything works now.

samestep (approved):

athas:
Looks good to me.

PR #102, opened 2024-10-29
GMM for TensorFlow

maggiehollis:
Implements [GMM](https://github.com/microsoft/ADBench/blob/master/src/python/modules/Tensorflow/TensorflowGMM.py) for TensorFlow


samestep (approved):
Looks great, thanks @maggiehollis!

PR #103, opened 2024-10-31
GMM for Autograd

maggiehollis:
Implements [GMM](https://github.com/microsoft/ADBench/tree/master/tools/Autograd) for Autograd

PR #110, opened 2024-11-05
BA for Tensorflow

maggiehollis:
Implementation of [BA](https://github.com/microsoft/ADBench/blob/master/src/python/modules/Tensorflow/TensorflowBA.py) for TensorFlow


samestep (changes requested):
Thanks for doing this! One step closer to parity with ADBench :)

Similar to #103, this implementation is failing the validation added in #101; see [the CI logs](https://github.com/gradbench/gradbench/actions/runs/11731154326/job/32681277625?pr=110). If you look at the full log, you can see that the reason the `assert a == b` check is failing is because the TensorFlow tool implementation is simply returning `"output": "Process terminated due to timeout."`; is there a way to fix this?

maggiehollis:
I reverted the merge from main because it presented a new issue with diffsharp?

athas:
That's a transient issue with DiffSharp, which is flaky. It's unrelated to any changes in the main branch.

samestep:
The commit c45de3903c8108d2d0a91ac4640630e5739653ec seems to have messed up the Git history so I'm force-pushing to 437b6de89521974f8b9469aeb554bf068827f0dc.

samestep (changes requested):
Seems to be working now that the timeout is removed! @maggiehollis would you be able to remove the commented-out timeout code? Then this should be ready to merge.

samestep (approved):
Looks good, thanks Maggie!

PR #126, opened 2024-11-22
Port HT from ADBench.

athas:
ADBench is a bit inconsistent about whether this benchmark is abbreviated "HT" (for Hand Tracking) or "HAND". I have chosen the former, as it is more consistent with the other names of ADBench benchmarks.

Currently two tools implement this eval: PyTorch (implementation also from ADBench) and Futhark. They validate against each other.

The run.py file for ht is currently hardcoded to only consider two datasets for each of the four "classes", because the larger datasets are extremely time consuming to run with PyTorch.


samestep (changes requested):
Look awesome! Still reviewing most of this, but while I do that, would it be possible for you to move the HT data into `evals/ht` instead of `python/gradbench/evals/ht`? Currently the entirety of the `python` directory gets included in every tool and eval that has any Python code, so I'd like to try to keep it smallish; the rest of the contents of that directory are 100x smaller than these new additions.

python/gradbench/adbench/ht_data.py
samestep, line 1: All the other files in this directory contain a permalink to their source in the [original ADBench repo](https://github.com/microsoft/ADBench), along with a comment listing any changes made; could you add that here too?

python/gradbench/evals/ht/io.py
samestep, line 1: Same request for this file, could you link to the original and list any changes made?

python/gradbench/pytorch/ht.py
samestep, line 1: Ditto here.

python/gradbench/pytorch/ht_objective.py
samestep, line 1: Here as well.

samestep (changes requested):
And just a few other minor requests. Other than that, looks great to me, thanks for doing the work on this!

samestep (approved):

PR #127, opened 2024-11-26
Allow command line options to be passed to tools and evals.

athas:
Also augments the ba/gmm/ht evals to accept command line options indicating which workloads to run. This is probably the most useful usage of command line options.

By default, when no options are provided, the current selection of workloads is preserved.


samestep (approved):
This is excellent, thank you!

PR #128, opened 2024-11-26
Extend protocol with an "analysis" message kind.

athas:
This message is sent by the eval after it receives a response to an evaluation message, and contains validation information. The idea is that this is used by the run.py script (or equivalent). The eval does not expect a response to the analysis message. The ID of the analysis message is the same as the ID of the evaluation message.

This PR does not change the contents of the "end" message, so validation information is now reported twice (but still only computed once, so there is little risk of inconsistency).


samestep (approved):
Looks good to me!

PR #129, opened 2024-11-26
Add workload labels.

athas:
This extends the eval protocol such that "evaluate" messages must now also have a "workload" field. This field must contain a human-readable string. It is intended for use by analysis tools and carries no semantic meaning. It is intended that every distinct "input" also has a distinct "workload" label.

### DIFF
--- a/python/gradbench/evals/ba/run.py
+++ b/python/gradbench/evals/ba/run.py
@@ -66,8 +66,8 @@ def main():
             datafile = next((Path(__file__).parent / "data").glob(f"ba{i}_*.txt"), None)
             if datafile:
                 input = parse(datafile)
-                e.evaluate(name="calculate_objectiveBA", input=input)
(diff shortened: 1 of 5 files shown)


samestep (changes requested):
Personally I'd rather the `"workload"` be optional; would that be OK?

athas:
What about defaulting the workload to a string representation of the ID in case it is not provided? That saves on control flow in tools that consume the log files.

samestep:
Personally I'd rather the `"workload"` be genuinely optional, and downstream tools do something like
```python
analysis.get("workload") or str(analysis["id"])
```
rather than doing that in `evaluation.py`; semantically, I'd prefer for "no workload name was provided" to be distinguishable from "the workload name happened to be the same as the message ID".

athas:
Done.

samestep (approved):
Thank you!

PR #131, opened 2024-11-26
Fix workload labels.

athas:
Not sure how that snuck through.


samestep:
Bizarre. Maybe some weird behavior of Black autoformatting? This is why I always have it run on save.

samestep (approved):

PR #134, opened 2024-11-29
Implement ADBench benchmarks for Tapenade.

athas:
This PR begins the work of including Tapenade in Gradbench. Currently only GMM is operational, but it does validate.

This work includes moving a lot of code from ADBench to Gradbench (and also a JSON parser for C++). Some of the scaffolding code may eventually be useful for other tools that are unfortunate enough to speak C++.

The actual Gradbench-specific code is contained in the run_gmm.cpp program.

Sadly, except for trivial evals (such as "hello"), running Tapenade itself at define-time seems unviable, as the output code must in all cases be manually modified in order to compile, and have various supporting structures defined in seemingly ad-hoc ways. I have simply, as with ADBench, included the thus-modified code.


athas:
BA should also validate now, although it's not much of an achievement as I remembered the BA implementations only report the size of the (sparse) Jacobian, not its contents. They might be wrong.

athas:
This should all work now, although the DiffSharp failure prevents the CI tests from running.

tools/tapenade/Dockerfile
```diff
@@ -5,18 +5,22 @@ WORKDIR /home/gradbench
...
 RUN tar -xvzf tapenade_3.16.tar
 
-
 # ensure Tapenade is accessible
 RUN ln -s /home/gradbench/tapenade_3.16/bin/tapenade /usr/local/bin/tapenade
 
+# Make the cpp directory available in a somewhat odd location.
+COPY cpp /cpp/
+
 COPY tools/tapenade/ .
```
samestep, line 21: I realize now that I forgot to include DiffSharp SciLean, and Tapenade when I did this (and also I didn't document this principle anywhere), but in #80 I established the convention that, when we `COPY` files from this repo into a Docker image, we preserve their relative locations under `/home/gradbench`; could we do the same here?

```suggestion
COPY cpp /home/gradbench/cpp
COPY tools/tapenade /home/gradbench/tools/tapenade
```

cpp/json.hpp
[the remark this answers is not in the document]
athas: Regarding json.hpp, I can do that, but it would make it impossible to build any of the C++ tools outside of Docker, which seemed to be desirable in some of the other PRs.

tools/tapenade/Dockerfile
athas: Sure, I don't think that should be a problem.

athas:
All of the files you identified as type 2 are correctly identified. The following files you identified as type 1 are type 2:

* tools/tapenade/hand/hand_d.c
* tools/tapenade/hand/hand_d_tapenade_generated.c

All of the files with the Microsoft notice are type 2. I simply missed adding the links because of the sheer amount of files added. You're right that it's unclear how specifically to annotate the Tapenade-generated, since they in principle should also contain the annotation from the *original file*. I don't think it's important as long as they contain *one* link.

I'll add the missing links.

tools/tapenade/Dockerfile
athas: Done.

athas:
I have added the links.

cpp/json.hpp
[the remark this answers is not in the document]
athas: This seems like a lot of machinery for a library that is explicitly designed to be vendored. There is no such widely used package downloading machinery for C++, unless you want to turn GradBench into a full cmake-based project including that terrible package manager they have (which ADBench uses, actually). Alternatively, we could include that JSON library as a Git submodule, but nobody remembers to clone with `--recursive`, so this would also just break mysteriously for people.

samestep (approved):
Made a few changes; I'm ready to merge this now if you think it looks good!

Regarding the JSON library, what are your thoughts on downloading it as I've done in df1118f5a916064506ed2005120bf3d373c55ef8?

samestep:
Yeah I agree that the added bit of complexity is annoying, but I do prefer this since it makes it crystal clear where the file comes from. For instance, before 581ce4a8383a05bf5037d838d744219045956e3d I actually thought the file did come from the 3.11.3 release; only after looking at the Git diff when running `wget` the first time did I realize that it actually came from a different commit from that repo.

PR #135, opened 2024-11-30
Futhark cleanup.

athas:
This moves all the Futhark-relevant code into tools/futhark, since it is not interesting to other tools.


samestep (approved):
Cool thanks! Made a couple additional changes to make it more convenient to run outside Docker.

PR #137, opened 2024-12-01
Add more human-friendly intermediary.

athas:


athas:
I realised this doesn't run without modifying the pyproject.toml to add termcolor. (My dirty secret is that I actually use a shell.nix for the Python dependencies.) I also realised that it might be pretty easy to modify this to optionally generate a JSON output file just like `run.py`, and in that case, why have two intermediaries at all?

samestep:
@athas yeah; as an aside, I've also been considering switching from Poetry to [uv](https://docs.astral.sh/uv/), but that's not really important. As I mentioned in https://github.com/gradbench/gradbench/issues/99#issuecomment-2512814744 though, I've been considering changing the output from the intermediary to just be [JSON Lines](https://jsonlines.org/) instead of trying to massage it into actual JSON, which would enable the intermediary to just pipe to another tool or use `tee` or something.

I've also considered switching the intermediary from being a random Python script to being part of a slightly larger set of convenience tools, which we'd let people install via pip or Cargo or something. So maybe there'd be subcommands like `gradbench run` but also `gradbench buildtool` to replace `buildtools.h`, or `gradbench summarize` to replace `.github/summarize.py`, or `gradbench serve` to start a webserver on `localhost` to view some data in a browser, etc.

athas:
I don't have a strong opinion, but I'll make the following points:

1. Most contributions from the AD community will (hopefully) be benchmark implementations/tools. A smaller amount of contributions will be new evals. Almost nobody will write new intermediaries. In fact, I think it is perfectly reasonable to have only a single intermediary, ever. This suggests that the most important issue is to make it simple to write new benchmark implementations (we don't actually have a term for this in the documentation).
2. JSON Lines is pretty much the format that evals and tools currently exchange, and I think it is very nice for that purpose. It means you don't need a streaming JSON parser, as you can start by reading a full line and then using a standard whole-input JSON parser.
3. GradBench does not have a term for a program that consumes one or more *complete* log file (like the one produced by `run.json`), and produces e.g. graphs of runtimes. In any case, I don't think it is important that these log files are in JSON Lines format, since I don't think such "analysis programs" (for lack of a term) will be super common, and they will be operating on complete JSON files anyway.
4. It is important that Gradbench comes with useful intermedaries and analysis programs by default, which should not require installing other tools, such as a Rust compiler. Such things can be dockerized, but that is sometimes awkward to use.
5. The current utility scripts (buildeval.sh, eval.sh. etc) are fine, but I found them a bit confusing at first, and so might others. Adding a nicer frontend (maybe having an intermediary automatically rebuild whatever is needed) would be useful. Similarly, some automation for running multiple tools and evals is also useful. Currently this has to be done manually (GA does it via the build matrix, but that's only useful in CI).

I think one of the most important next steps (after finishing the current PRs) is actually to implement an example of a point 3 program, so we will know whether the current foundation is solid.

samestep (approved):
Looks good, thanks for publishing this script! Let's go ahead and merge this, and I'll work on the unified frontend idea as a followup.

PR #136, opened 2024-12-01
Port Manual and Finite implementations from ADBench

athas:
This PR ports the manually differentiated and finite differences implementations from ADBench.

It is marked as a draft as it depends on (and currently includes) the infrastructure added in #134. I intend to rebase it once the main branch has C++ infrastructure.

The Finite tool does not validate, due to roundoff errors. We will need to adjust validation to use a greater tolerance.

PR #138, opened 2024-12-02
Increase tolerance when validating gmm and hello.

athas:
The tolerances here are somewhat arbitrarily set to allow validation with an implementation based on finite differencing. There is room to bikeshed them, but I suggest not doing so, for the following reasons:

  0) Even with the very generous gmm tolerance, it is unlikely that a
  buggy implementation will have a validating Jacobian.

  1) Although a loose tolerance allows implementations to play
  mixed-precision games for performance, Gradbench is not a
  competition, and so we can assume "good faith" on behalf of the
  tools.

Options for configuring the tolerance can always be added later.

### DIFF
--- a/evals/hello/Dockerfile
+++ b/evals/hello/Dockerfile
@@ -1,6 +1,6 @@
 FROM python:3.11-alpine
 
-RUN pip install pydantic
+RUN pip install numpy==1.26.0 pydantic
 
 COPY python /home/gradbench/python
 ENV PYTHONPATH=/home/gradbench/python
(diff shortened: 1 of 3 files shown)


samestep (approved):
Fine with me! Out of curiosity, did you run into issues with these tolerances before this PR, or is this just a precaution for the future?

athas:
The finite differences implementation in #136 doesn't validate without these relaxed tolerances. I expect finite differences will be the numerically worst of any tool for all benchmarks (unless someone is doing something seriously funky), so I don't expect and endless churn of fiddling with tolerances.

PR #145, opened 2024-12-06
Use gmm workloads with different values of d.

athas:
I had to modify the data generator to get this to work.


athas:
Debugging the tensorflow failure will be tricky, as the log file is so large that GitHub will not even let me download it. I suspect it may be time to not print all the inputs and results in the log.

My hunch is that the tensorflow failure is due to OOM, but I ran it locally and it doesn't look excessive to me.

samestep (approved):
If CI passes then this looks good to me!

PR #146, opened 2024-12-07
Add --json option to run-pretty.py.

athas:
With this change, there is strictly speaking no reason to still keep run.py around, unless you really like enormous amounts of JSON in your terminal.


athas:
Closed in favour of #147.

PR #147, opened 2024-12-08
Consolidate scripts into one unified CLI

samestep:
Following up on https://github.com/gradbench/gradbench/pull/137#issuecomment-2512830348, this PR replaces many shell and Python scripts with a single `gradbench` CLI, with the following benefits:

- easy to install and use without cloning the repo
- automatic usage validation
- better documentation
- individual subcommands are more discoverable

This supercedes #146. (As a side note, that PR is not strictly correct according to the protocol we currently have documented, as it omits the top-level `"nanoseconds"` from each round of the log.)

A couple followups that would be nice to do after this PR:

- publish binaries for popular platforms to allow installation without having Rust installed
- revamp the protocol to simplify it and output the log as [JSON Lines](https://jsonlines.org/) instead of just JSON


README.md
```diff
@@ -20,17 +20,16 @@ See https://gradben.ch for a daily overview of all the tools (columns) and bench
 
 ## Usage
 
-If you have [Git][] installed, you can clone this repository; for instance, via the [GitHub CLI][]:
+If you have [Rust][] installed, you can download and install the GradBench CLI:
 
 ```sh
-gh repo clone gradbench/gradbench
-cd gradbench
+cargo install --locked --git https://github.com/gradbench/gradbench
```
athas, line 26: It is odd to install the `gradbench` CLI like this, since anyone actually using GradBench will presumably have a checkout of the repository. Apart from putting binaries outside of the GradBench git checkout itself, this also runs the risk of people using an outdated `gradbench` CLI. There is a very low chance of actually getting the `gradbench` CLI into any package repositories, so the documentation and the architecture should be written such that this is not the expected outcome.

The instructions in CONTRIBUTING.md are a better workflow.

README.md
```diff
@@ -20,17 +20,16 @@ See https://gradben.ch for a daily overview of all the tools (columns) and bench
...
 ```
 
-Then if you have [Python][] and [Docker][] installed, you can run any of our [evals](evals) against any of our [tools](tools) via our `run.py` script:
+Then if you have [Docker][] installed, you can use the GradBench CLI to run any of our [evals](evals) against any of our [tools](tools):
 
 ```sh
-./run.py --eval './eval.sh hello' --tool './tool.sh pytorch'
+gradbench run --eval 'gradbench eval hello' --tool 'gradbench tool pytorch'
```
athas, line 32: Use `./gradbench` here, just as in CONTRIBUTING.md.

Actually I would just move the material from CONTRIBUTING.md here.

athas (changes requested):

samestep:
@athas To me, by convention, `README.md` is for explaining what the project is about and how to use it without cloning the repo, and `CONTRIBUTING.md` is for explaining how to work on the project itself, usually after cloning the repo. So I intend to leave `README.md` with a link to https://gradben.ch and instructions on how to install and use `gradbench`, and then use `./gradbench` in `CONTRIBUTING.md`.

If you think I should move the section in `README.md` that _links_ to `CONTRIBUTING.md` to be closer to the top rather than at the bottom, though, I'd be happy to move that.

athas:
Who would ever use GradBench without having cloned the repo? (Or working in a tarballed copy of it.) I would be very mystified by the description, as it is unlikely any benchmark suite I have seen before (while the reality is that GradBench is structured much like they all are).

samestep:
The two commands in `README.md` in this branch work perfectly well without cloning the repo, since all GradBench nightly Docker images are published to the GitHub Container registry.

README.md
```diff
@@ -20,17 +20,16 @@ See https://gradben.ch for a daily overview of all the tools (columns) and bench
...
 ```
 
 This will first automatically download our latest nightly Docker images for the given eval and tool, and then run the eval against the tool while printing the entire communication log to the terminal.
```
athas, line 35: The entire communication log is not printed to the terminal anymore.

samestep: Right, thanks for the catch! Fixed.

athas:
If I'm modifying or adding an eval or a tool (which is the only reason I can think of for running GradBench yourself), then they won't work. But if you think they're useful, then we can leave those instructions. I think it's unfortunate that the README will contain commands that will not work unless you install `gradbench` in your PATH, however.

The actual CLI tool is good and seems for work for everything I've tried. To bikeshed a bit, perhaps `gradbench repo tool` should be `gradbench repo buildtool` (and similarly for `eval`), so it doesn't look so similar to 'gradbench tool' at a glance.

samestep:
Since `gradbench run` lets you give any command (it's decoupled from `gradbench eval` and gradbench tool`), you still can directly use it even when adding an eval or a tool. It'd be pretty much fine to use it even when working in a clone of the GradBench repo too, since the set of actual evals and tools is not baked into the `gradbench` binary at all.

Renaming `repo eval` and `repo tool` to `repo buildeval` and `repo buildtool` respectively sounds good to me, I'll make that change.

athas (approved):

PR #148, opened 2024-12-08
Implement hello eval with Enzyme.

athas:
I'm mostly interesting in feedback on the Dockerfile/Makefile setup. In principle supporting the remaining evals should be straightforward.

Updating the version of Enzyme in the Dockerfile should not be difficult, assuming they don't change the build instructions or add interesting new dependencies.

I think the C++ infrastructure ported from ADBench really shows its advantages here. Perhaps something similar should be done for all the trivial Python wrappers - they are *very* similar.


tools/enzyme/gmm.py
```diff
@@ -0,0 +1,27 @@
+import json
+import subprocess
+import tempfile
+from os import listdir
+
+
+def compile():
+    # Nothing to do here. We assume everything is precompiled.
+    return True
```
samestep, line 9: Ideally we'd actually run Enzyme on receiving a `"define"` message. Is this infeasible for similar reasons to Tapenade? Or would it actually be possible to do with Enzyme? I'm really interested in being able to do this, since I've heard that Enzyme is slow and it'd be nice for GradBench to be able to quantify that.

samestep:
Marking as draft since [the Enzyme tool is claiming to support GMM but actually fails on it](https://github.com/gradbench/gradbench/actions/runs/12225471023/job/34104456321?pr=148); feel free to undraft once that's no longer the case.

tools/enzyme/gmm.py
athas: No, this should be fine, actually.

athas: Done now; although there's room to tune what exactly is measured. I think Enzyme is only a small part of the whole C++ compilation process.

samestep: Awesome, thanks! Yeah I think after we merge #149 (assuming it looks OK to you) we should add an optional `"timings"` field to the tool's response to `"define"` messages, which would have the same format as for `"evaluate"` (a list of `"name"`/`"nanoseconds"` pairs).

samestep (changes requested):
I'd rather delete all the GMM code in this PR before merging, since if it is of interest then it can be seen by looking at the Git history of this PR.

athas:
GMM has been purged.

samestep (approved):
Looks good to me, thanks for getting this working!

PR #151, opened 2024-12-09
Support multiple runs.

athas:
This currently changes only the gmm eval and the Futhark tool.

### DIFF
--- a/tools/futhark/run.py
+++ b/tools/futhark/run.py
@@ -29,11 +29,13 @@ def run(params):
     with futhark_server.Server(server_prog(params["module"])) as server:
         prepare = resolve(params["module"], "prepare")
         run = resolve(params["module"], params["function"])
-        prepare(server, params)
-        start = time.perf_counter_ns()
-        ret = run(server)
-        end = time.perf_counter_ns()
-        timings = [{"name": "evaluate", "nanoseconds": end - start}]
+        prepare(server, params["input"])
+        if "runs" in params["input"]:
+            runs = params["input"]["runs"]
+        else:
+            runs = 1
+        ret, times = run(server, runs)
+        timings = [{"name": "evaluate", "nanoseconds": ns} for ns in times]
         return {"output": ret, "timings": timings}
 
 
--- a/python/gradbench/evals/gmm/run.py
+++ b/python/gradbench/evals/gmm/run.py
@@ -20,6 +20,7 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("-n", nargs="+", type=int, default=[1000, 10000])
     parser.add_argument("-k", nargs="+", type=int, default=[5, 10, 25, 50, 100, 200])
+    parser.add_argument("--runs", type=int, default=10)
     args = parser.parse_args()
 
     e = SingleModuleValidatedEvaluation(module="gmm", validator=assertion(check))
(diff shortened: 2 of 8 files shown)


samestep:
Marking as draft for now since CI is currently failing.

athas:
This is designed to be minimally invasive and strictly speaking does not require any protocol extension. It also turns out the generated `from_dict` methods are lenient regarding bonus fields.

samestep:
@athas could you fix the merge conflict?

tools/futhark/run.py
```diff
@@ -29,11 +29,13 @@ def run(params):
...
-        start = time.perf_counter_ns()
-        ret = run(server)
-        end = time.perf_counter_ns()
-        timings = [{"name": "evaluate", "nanoseconds": end - start}]
+        prepare(server, params["input"])
+        if type(params["input"]) is dict and "runs" in params["input"]:
+            runs = params["input"]["runs"]
+        else:
+            runs = 1
+        ret, times = run(server, runs)
```
samestep, line 37: This is awkward and fragile, since rather than conservatively using just the known semantics from the evals that currently exist, it assumes that a `"runs"` field in the `"input"` will have the same semantics in all future evals.

Actually, I didn't realize until just now that all the Futhark implementations assume that all the functions in a module take the exact same inputs as each other, which is not guaranteed by the protocol. It seems to me that it'd be better to attach the `prepare` function to the individual functions themselves, e.g. via the `wrap` decorator available in `python/gradbench/wrap_module.py`, already used in the GradBench implementations for Autograd, JAX, PyTorch, and Tensorflow:

https://github.com/gradbench/gradbench/blob/9c3950d7ee034d534db51658a6aebdf855cc1d23/python/gradbench/pytorch/gmm.py#L79-L102

Can we do the same thing here? Specifically, this can be decomposed into two changes:

1. Have the `prepare` function directly return anything that needs to be passed to the `run` function; in this case, it looks like that's just the number of `"runs"`.
2. Rather than implicitly using a function named `prepare` from the containing module, instead use `@wrap` (or something similar) to attach it to to each individual function. The `prepare` function can still be defined just once if it is indeed shared, of course.

samestep (changes requested):

tools/futhark/run.py
athas: That seems orthogonal to this PR. It can be easily fixed if/when it becomes an issue. Arguably, there is no need for Futhark to even have those `prepare` methods - uploading the input data can just be put directly into the run methods.

samestep: You're right that the second proposed change is pretty orthogonal to this PR, so I'm happy to do it in a followup. I don't see the first as orthogonal, since this PR introduced that new `if`/`else` branch in `tools/futhark/run.py`, so I'd prefer to remove that before merging.

athas: Removing that branch (rather than just moving it elsewhere) requires modifying the `"input"` field of the `hello` eval to be a dict like the others, which will require modifying every tool. Do you want me to do that in this PR?

samestep: No, that's neither desirable nor required. I've removed the branch in ea279948ab3d7410a19656835830d1676d44075d. Does this look reasonable?

samestep (approved):
This looks good to me if you're happy with it!

PR #158, opened 2024-12-10
Make ba implementations produce actual Jacobian.

athas:


samestep (approved):
Looks good to me, thanks!

PR #162, opened 2024-12-11
Add --runs to all ADBench-derived evals.

athas:
Also extends the PyTorch and Futhark tools to respect runs.

I had to perform some surgery on how timings are performed. Now, rather than doing it in the generic run.py module itself, it is done in each of the tool-specific eval modules. I think that is ultimately the best place to do it, since it depends on the specific eval what is to be measured, and how. The run function now returns both the result and a list of timings (the latter which is just put into the response message).


athas:
I also removed the timeout code from ba.py since it got in the way of my regularisation of the wrapping/unwrapping machinery. I suspect timeouts are better handled in the general run.py for the tool, rather than having an ad-hoc implementation per eval.

samestep (changes requested):
Still looking through the code in detail, but in the meantime, could you change the `default` for `--runs` to be just `1` instead of `10`? Per our [prior](https://discord.com/channels/1242864568850579487/1267596255886970952/1314686303874453505) [conversation](https://discord.com/channels/1242864568850579487/1267596255886970952/1314686402667090102) on the Discord:

> for now, i'd probably say let's leave the nightly the same as the pr ci, but if any evals are missing flags to configure how big their workload is, we can definitely add those

> and then later once we set up self-hosted runners or something to get better numbers from nightly, we can bump up those workloads to something more substantial

Looking back at those messages, I guess I didn't explicitly say I'd rather just have `--runs=1` by default for now, but that was what I meant to say. I mostly just don't want PR authors to have to wait [over an hour](https://github.com/gradbench/gradbench/actions/runs/12274893466/job/34257376465) to get validation results.

athas:
Yes, I agree. I was actually going to suggest that myself.

samestep:
@athas I made some changes in f62c478e8b87d8ff97982dff61d44daa5da8d8f4; specifically, I wasn't a fan of the following:

- It looked like you were actually looping over `runs` _twice_, resulting in `runs*runs` instead of just `runs` (unless I was missing something).
- The logic for timing one or more runs of a function became duplicated in many more places.
- Returning just a tuple seemed like not the greatest interface.

So I renamed `python/gradbench/wrap_module.py` to `python/gradbench/wrap.py`, renamed the wrapped `Functions` type to just `Wrapped` and cleaned it up a bit, renamed `@wrap` to `@wrap.function` which now gives a `wrapped` function that returns `OutputAndTimings`, and also added a new `@wrap.multiple_runs` decorator to make it easy to support `--runs`.

Does this look reasonable?

athas:
The `.prepare` methods might be expensive. This is why I took care to put them outside of the timing loops.

samestep:
@athas sure, but that was already the case before this PR, right? Could we just do the "handle multiple `--runs` stuff in this PR, and then do the "make the scope of the `"evaluate"` timing" stuff in a followup?

athas:
Only GMM did multiple runs before this PR. Why add more buggy features that must be fixed immediately? We might as well get it right this time around, especially as we're also adding/modifying an abstraction.

athas:
That said, I think the `multiple_runs` abstraction is fine in the long run. It's doing the initialisation in the `prepare_input` method that really matters to me, and that's still possible.

samestep:
What new buggy features are you referring to? All this does is make PyTorch support `--runs`, and improves our decorator abstraction. I don't see any bugs this adds that weren't already there?

athas:
We're adding incorrect timing loops to more programs. The fact that GMM already had an incorrect timing loop is I think immaterial. We might as well get it right now to save ourselves another PR, especially since I already had corrected versions of the input preparation methods in a previous commit (plus some fiddling with the times that you found a better solution for).

Regarding my remark about abstractions, I was uncertain at first whether this form of `wrap` made an assumption that would make it impossible to lift the input preparation outside of the timing loop, but it is now clear to me that is not the case.

samestep:
I don't quite follow, but I don't think it really matters; since it seems like we agree that the new abstraction is OK and enables multiple runs, can we just go ahead and merge this PR (once CI passes) and then do what you're proposing immediately after?

samestep (approved):

athas:
Sure.

PR #164, opened 2024-12-12
Document HT and LSTM.

athas:


athas:
Note that I invented types `int` and `double` for the specification, because it is really critical that some things are integers, and TypeScript apparently cannot express this.

samestep (approved):
Awesome thanks! These seem like a great start, and we can continue to improve documentation more going forward.

PR #165, opened 2024-12-12
Construct the ADBench wrapper objects in prepare_input method.

athas:
Importantly, this moves the invocation of the "prepare" method out of the timing loop. Depending on the PyTorch backend, "prepare" might be very costly (e.g. copying data to a GPU), but is unrelated to AD.

A future addition could be to also return timing information for "prepare", in case anyone would find that interesting.

### DIFF
--- a/python/gradbench/pytorch/ba.py
+++ b/python/gradbench/pytorch/ba.py
@@ -122,10 +122,6 @@ class PyTorchBA(ITest):
             self.reproj_error = reproj_error.flatten()
 
 
-def evaluate_times(times):
-    return [{"name": "evaluate", "nanoseconds": time} for time in times]
-
-
 # Convert objective output to dictionary
 def objective_output(errors):
     r_err, w_err = errors
@@ -171,20 +167,18 @@ def prepare_input(input):
         camIdx = (camIdx + 1) % n
         ptIdx = (ptIdx + 1) % m
 
-    return BAInput(cams, X, w, obs, feats)
+    py = PyTorchBA()
+    py.prepare(BAInput(cams, X, w, obs, feats))
+    return py
 
 
 @wrap.multiple_runs(runs=lambda x: x["runs"], pre=prepare_input, post=objective_output)
-def objective(input):
-    py = PyTorchBA()
-    py.prepare(input)
+def objective(py):
     py.calculate_objective(1)
     return py.reproj_error, py.w_err
 
 
 @wrap.multiple_runs(runs=lambda x: x["runs"], pre=prepare_input, post=jacobian_output)
-def jacobian(input):
-    py = PyTorchBA()
-    py.prepare(input)
+def jacobian(py):
     py.calculate_jacobian(1)
     return py.jacobian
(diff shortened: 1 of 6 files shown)


samestep:
Does this actually affect the timings significantly? What are the numbers?

athas:
I don't know how to set up PyTorch with passthrough for my AMD GPU (which is all I have here), so I can't measure the specific impact, but it's pretty standard practice to at least isolate data movement costs. But on a CPU, here's the first few GMM runtimes on `main` with 100 runs:

```
  [0] start
  [1] def   gmm                                 1.532 s ✓
  [2] eval  gmm::objective  2_5_1000              123ms ~       120ms evaluate×100 ✓
  [4] eval  gmm::jacobian   2_5_1000              201ms ~       200ms evaluate×100 ✓
  [6] eval  gmm::objective  2_10_1000             143ms ~       142ms evaluate×100 ✓
  [8] eval  gmm::jacobian   2_10_1000             246ms ~       244ms evaluate×100 ✓
 [10] eval  gmm::objective  2_25_1000             148ms ~       147ms evaluate×100 ✓
 [12] eval  gmm::jacobian   2_25_1000             350ms ~       348ms evaluate×100 ✓
 [14] eval  gmm::objective  2_50_1000             173ms ~       171ms evaluate×100 ✓
 [16] eval  gmm::jacobian   2_50_1000             459ms ~       457ms evaluate×100 ✓
 [18] eval  gmm::objective  10_5_1000             186ms ~       181ms evaluate×100 ✓
 [20] eval  gmm::jacobian   10_5_1000             385ms ~       380ms evaluate×100 ✓
```

And here's with this change:

```
  [0] start
  [1] def   gmm                                 1.538 s ✓
  [2] eval  gmm::objective  2_5_1000               92ms ~        88ms evaluate×100 ✓
  [4] eval  gmm::jacobian   2_5_1000              174ms ~       172ms evaluate×100 ✓
  [6] eval  gmm::objective  2_10_1000             109ms ~       107ms evaluate×100 ✓
  [8] eval  gmm::jacobian   2_10_1000             218ms ~       216ms evaluate×100 ✓
 [10] eval  gmm::objective  2_25_1000             118ms ~       116ms evaluate×100 ✓
 [12] eval  gmm::jacobian   2_25_1000             290ms ~       287ms evaluate×100 ✓
 [14] eval  gmm::objective  2_50_1000             150ms ~       148ms evaluate×100 ✓
 [16] eval  gmm::jacobian   2_50_1000             366ms ~       363ms evaluate×100 ✓
 [18] eval  gmm::objective  10_5_1000             100ms ~        94ms evaluate×100 ✓
 [20] eval  gmm::jacobian   10_5_1000             276ms ~       270ms evaluate×100 ✓
```

samestep (changes requested):
Seems reasonable to me. The only thing I was a bit unsure about is whether the `calculate_objective` or `calculate_jacobian` methods are too stateful somehow, but since it validates, seems fine.

Could you do the same for TensorFlow, for consistency?

athas:
I will do that.

For any code ported from ADBench, naive repeated application will work, since their implementations adhere to a principle of completely overwriting the output data structure.

samestep (approved):

PR #166, opened 2024-12-13
Support multiple runs in C++-based tools.

athas:
Rather than extending the ADBench-defined input data structures with a "runs" field, I opted to make the JSON parsing functions (which are custom to GradBench) write to a separate "runs" variable. This was done solely to minimise code differences with ADBench.

### DIFF
--- a/cpp/adbench/io.h
+++ b/cpp/adbench/io.h
@@ -7,22 +7,22 @@
 #include "adbench/shared/LSTMData.h"
 #include "adbench/shared/HTData.h"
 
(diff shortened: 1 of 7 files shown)


cpp/adbench/io.h
samestep, line 1: Instead of just `int*`, can you write a parameter name like `int* runs`? (since all the other parameters here have names)

samestep (changes requested):
Seems reasonable overall! Just one request.

samestep (approved):

PR #167, opened 2024-12-13
Make objective_hello a template function.

athas:
This is to ensure that it is usable with operator overloading-based AD tools, which is consistent with the other C++ objective functions.

Of course the 'hello' eval is so trivial that it takes less time to implement than to import, but I think it is useful to use it for demonstrating "best practice".


samestep (approved):
Seems straightforward to me, thanks!

PR #171, opened 2024-12-16
Add ADOL-C tool.

athas:
Will eventually port ADBench implementations, but hello is implemented from scratch.

ADOL-C has a very weird API.

### DIFF
--- a/tools/adol-c/.gitignore
+++ b/tools/adol-c/.gitignore
@@ -1,0 +1,6 @@
+*.o
+run_hello
+run_gmm
+run_ba
+run_lstm
+run_ht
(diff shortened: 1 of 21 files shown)


athas:
This now compiles at run-time and I have removed the copyright notices.

athas:
It's worth noting that most of the precompilation time is probably spent on compiling stuff like the `utils.cpp` file, so I'm not sure the numbers are particularly interesting. But now they're there.

samestep:
Yeah that is a bit unfortunate; hopefully with subtask timings we'll be able to separate out the interesting things, We'll see.

samestep (approved):

PR #174, opened 2024-12-18
Deduplicate C++ run.py code.

athas:
Also shifts responsibility for producing JSON timings to the C++ tool implementations. In practice, this is done by the shared main.h file. This can be used to provide measurements of sub-components of the evaluation, such as the time taken by the 'prepare' method.


samestep (approved):
Cool! Looks good to me, thanks for this nice refactor.

Issue #139, opened 2024-12-02
Summarize validation table on pull requests


samestep:
Currently the https://gradben.ch website can only show data from nightly runs. It'd be nice if maybe it could also be used to show data from pull requests, e.g. to quickly and easily check for basic regressions. Or maybe that data could just be presented in a Markdown table in the [job summary](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#adding-a-job-summary).

samestep:
See also #175: ideally this would also comment on the PR to list any _differences_ in this high-level table compared to `main`, to make it easy to tell if a PR accidentally causes a tool to stop supporting an eval.

samestep:
This situation is improved a bit by #247.

PR #178, opened 2025-01-07
Support timeouts in intermediary.

athas:
This does work, although it has the following deficiencies:

1. Timeouts are only supported for tool responses, not eval responses. This is probably not a significant problem.

2. Failures due to timeouts are not distinguishable from other failures. It can be expected that some tools will fail to complete on some datasets, and it would be nice if this did not cause CI to fail entirely, but simply show up on the graphs (once we get those). I think the solution here is to make the exit code depend on the reason for the termination. Alternatively, just don't treat timeouts as an error at all.

3. When run in the terminal, timeouts are somewhat noisy, because pipes get broken and the tool/eval scripts get upset about unexpected EOFs and SIGPIPEs and print out stack traces. Quite similar to what happens when a tool crashes, actually.

### DIFF
--- a/crates/gradbench/src/cli.rs
+++ b/crates/gradbench/src/cli.rs
@@ -11,6 +11,8 @@ use anyhow::anyhow;
 use clap::{Parser, Subcommand};
 use colored::Colorize;
 use serde::{Deserialize, Serialize};
+use std::time::Duration;
+use timeout_readwrite::TimeoutReader;
 
 /// CLI utilities for GradBench, a benchmark suite for differentiable programming across languages
 /// and domains.
@@ -75,6 +77,10 @@ enum Commands {
         /// A path to save the full log. For example: `log.jsonl`
         #[clap(short, long)]
         output: Option<PathBuf>,
+
+        /// The timeout for tool responses.
+        #[clap(long, default_value_t = 3600)]
+        timeout: u64,
     },
 
     /// Perform a task in a clone of the https://github.com/gradbench/gradbench repository.
@@ -340,12 +346,17 @@ fn print_status(success: bool) {
 }
 
 /// Run an eval and a tool together, returning the number of validation failures.
-fn intermediary(o: &mut impl Write, eval: &mut Child, tool: &mut Child) -> anyhow::Result<usize> {
+fn intermediary(
+    o: &mut impl Write,
+    eval: &mut Child,
+    tool: &mut Child,
+    timeout: Duration,
+) -> anyhow::Result<usize> {
     let mut invalid = 0;
     let mut eval_in = eval.stdin.take().unwrap();
     let mut tool_in = tool.stdin.take().unwrap();
     let mut eval_out = io::BufReader::new(eval.stdout.take().unwrap());
-    let mut tool_out = io::BufReader::new(tool.stdout.take().unwrap());
+    let mut tool_out = io::BufReader::new(TimeoutReader::new(tool.stdout.take().unwrap(), timeout));
     let start = Instant::now();
     let mut line = Line::new();
     while let Some(eval_line) = {
@@ -413,7 +424,21 @@ fn intermediary(o: &mut impl Write, eval: &mut Child, tool: &mut Child) -> anyho
         }
         io::stdout().flush()?;
         let mut tool_line = String::new();
-        tool_out.read_line(&mut tool_line)?;
+        if let Err(err) = tool_out.read_line(&mut tool_line) {
+            if err.kind() == io::ErrorKind::TimedOut {
+                writeln!(
+                    o,
+                    r#"{{ "elapsed": {{ "nanoseconds": {} }}, "response": "timeout" }}"#,
+                    (Instant::now() - start).as_nanos(),
+                )?
+            };
+            eval.kill()?;
+            tool.kill()?;
+            return Err(anyhow!(
+                "tool failed to respond within {}s",
+                timeout.as_secs()
+            ));
+        }
         let response_time = Instant::now();
         let nanos = (response_time - message_time).as_nanos();
         writeln!(
@@ -620,7 +645,12 @@ fn cli_result() -> Result<(), ExitCode> {
                 .arg(format!("ghcr.io/gradbench/tool-{tool}:{t}"))
                 .args(args))
         }
-        Commands::Run { eval, tool, output } => {
+        Commands::Run {
+            eval,
+            tool,
+            output,
+            timeout,
+        } => {
             let (Ok(mut client), Ok(mut server)) = (
                 Command::new("sh")
                     .arg("-c")
@@ -638,12 +668,13 @@ fn cli_result() -> Result<(), ExitCode> {
                 eprintln!("error starting eval and tool commands");
                 return Err(ExitCode::FAILURE);
             };
+            let timeout = Duration::new(timeout, 0);
             let result = match output {
                 Some(path) => match fs::File::create(&path) {
-                    Ok(mut file) => intermediary(&mut file, &mut client, &mut server),
+                    Ok(mut file) => intermediary(&mut file, &mut client, &mut server, timeout),
                     Err(err) => Err(anyhow!(err)),
                 },
-                None => intermediary(&mut io::sink(), &mut client, &mut server),
+                None => intermediary(&mut io::sink(), &mut client, &mut server, timeout),
             };
             match (&result, client.wait(), server.wait()) {
                 (&Ok(invalid), Ok(e), Ok(s)) => {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,6 +79,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ace50bade8e6234aa140d9a2f552bbee1db4d353f69b8217bc503490fc1a9f26"
 
 [[package]]
+name = "bitflags"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


athas:
I decided that timeouts on their own are not a reason to have a nonzero exit code. It is expected behaviour that many tools will time out for larger datasets, and this is *fine*. Only validation failure (or crashes) now result in nonzero exit codes.

This means that a broken change that makes a tool so slow that it times out for everything will not be noticed by CI. However, this is just a special case of making a tool unintentionally slower, which is already not caught by any automation. Automation that solves this latter general problem (#139) will also solve the specific case of a change making more evaluations time out.

crates/gradbench/src/cli.rs
```diff
@@ -638,17 +668,22 @@ fn cli_result() -> Result<(), ExitCode> {
...
             match (&result, client.wait(), server.wait()) {
-                (&Ok(invalid), Ok(e), Ok(s)) => {
+                (&Ok((invalid, timeout)), Ok(e), Ok(s)) => {
                     status_code(e)?;
-                    status_code(s)?;
+                    // In case of timeout, we don't expect the tool to
+                    // have died gracefully.
+                    if !timeout {
+                        status_code(s)?;
+                    }
```
samestep, line 686: Can you explain the reasoning here? If the tool timed out then we should be giving a nonzero error code anyways; why special-case this?

crates/gradbench/src/cli.rs
```diff
@@ -413,7 +425,20 @@ fn intermediary(o: &mut impl Write, eval: &mut Child, tool: &mut Child) -> anyho
         }
         io::stdout().flush()?;
         let mut tool_line = String::new();
-        tool_out.read_line(&mut tool_line)?;
+        if let Err(err) = tool_out.read_line(&mut tool_line) {
+            if err.kind() == io::ErrorKind::TimedOut {
+                let ns = (Instant::now() - start).as_nanos();
+                writeln!(
+                    o,
+                    r#"{{ "elapsed": {{ "nanoseconds": {} }}, "response": "timeout" }}"#,
```
samestep, line 433:
```suggestion
                    r#"{{ "elapsed": {{ "nanoseconds": {} }}, "timeout": true }}"#,
```

crates/gradbench/src/cli.rs
samestep: Never mind; somehow I missed your message above.

crates/gradbench/src/cli.rs
```diff
@@ -225,6 +230,18 @@ enum Message {
     },
 }
 
+impl Message {
+    /// Retrieve the `Id` from a message.
+    fn id(&self) -> Id {
+        match *self {
+            Message::Start { id } => id,
+            Message::Define { id, .. } => id,
+            Message::Analysis { of, .. } => of,
```
samestep, line 239: This is wrong, yeah? An analysis message still has an `"id"` which is specifically different from the ID of the message it's an analysis `"of"`. Actually, I just realized that I forgot to write that in the protocol documentation, so I'll do that as well.

crates/gradbench/src/cli.rs
```diff
@@ -75,6 +76,10 @@ enum Commands {
         /// A path to save the full log. For example: `log.jsonl`
         #[clap(short, long)]
         output: Option<PathBuf>,
+
+        /// The timeout in seconds for tool responses.
+        #[clap(long, default_value_t = 3600)]
```
samestep, line 81: Nice, somehow I didn't know about [`default_value_t`](https://docs.rs/clap/4.5.23/clap/_derive/index.html#arg-attributes). For consistency I'll also go do this on the other args that have defaults.

athas:
1. I can fix that.
2. The large stack trace is because I apparently overlooked `python/gradbench/evals/ba/run.py` in #179.
3. This is more difficult to fix. I *think* it is printed by the command `docker` itself, or some background process that it has launched.

athas:
Actually, it seems like the delayed message comes from `gradbench`, as this doesn't produce it:

```
./gradbench run --timeout 1 --eval './gradbench eval gmm' --tool 'docker run --rm --interactive ghcr.io/gradbench/tool-pytorch:latest'
```

The *real* underlying issue is actually that killing the tool process is usually pointless, as it's often just a shell running `gradbench` which is then running the tool. I think the right solution is to create a new Unix process group for the tool (and eval for that matter) so we can kill the process and all of its children. Otherwise it's not even a robust way to kill a runaway tool in the first place.

samestep:
@athas I'm coming back to this today, sorry for the delay.

> It seems like the delayed message comes from `gradbench`, as this doesn't produce it:
> 
> ```
> ./gradbench run --timeout 1 --eval './gradbench eval gmm' --tool 'docker run --rm --interactive ghcr.io/gradbench/tool-pytorch:latest'
> ```
> 
> The _real_ underlying issue is actually that killing the tool process is usually pointless, as it's often just a shell running `gradbench` which is then running the tool. I think the right solution is to create a new Unix process group for the tool (and eval for that matter) so we can kill the process and all of its children. Otherwise it's not even a robust way to kill a runaway tool in the first place.

Interesting. So, do you happen to know if this due to `gradbench eval` or `gradbench tool` doing something wrong in how they handle running their respective `docker` command, or due to `gradbench run` handling its two spawned processes wrong, or something else? In #147 I deleted `run-pretty.py`; did that script previously handle these sorts of things correctly? I tried to migrate from the Python APIs to the Rust APIs as best I could, but perhaps I did it subtly wrong.

If you're not sure off the top of your head then I can try to dig more into this.

athas:
To clarify further: the "broken pipe" message is emitted by `docker run`. From my explorations in cyberspace, it appears there is no way to avoid this. If `docker run` writes to a closed pipe, then it will emit that message. Hence the only way to avoid that message is to ensure that the `docker` process is killed before the pipe is closed. This is what is currently not done, unless the `docker` process is the *immediate* child of `gradbench`.

samestep:
OK, after playing around with it some more, I think I now understand the process group stuff you were talking about. I've pushed a commit that seems to handle that in the `--timeout` case, but it still doesn't correctly handle the <kbd>Ctrl</kbd>+<kbd>C</kbd> case. I'm going to try to figure that out next.

athas:
For Ctrl-c, you need to catch and handle SIGINT.

samestep:
OK @athas I just pushed commit afcd617328973156fedfd7274cc77073ee43d3f5 which updates this branch to use the new unit testing and error-reporting stuff I added in #226 and #228 respectively. This does pretty much everything as before, with a couple of differences:

- I didn't port the `killpg` code because it would have been a bit annoying to fit into the new code structure, and the output seems fine without it; example:

  ```
  $ ./gradbench run --timeout 2 --eval './gradbench eval ba' --tool './gradbench tool pytorch'
    [0] start
    [1] def   ba                                  1.094 s ✓
    [2] eval  ba::objective   ba1_n49_m777...     2.001 s ⧖
  write /dev/stdout: broken pipe
  ```

- I didn't port the code to emit a "synthetic timeout message" because that exception in the protocol seems a bit messy to me; I'd prefer to remove those modifications that this PR currently has in `README.md`. Could you remind me why we'd want this exception to the protocol?

Let me know your thoughts.

Also, as a separate topic, here is the new logic which says that timeouts don't produce a nonzero exit code:

https://github.com/gradbench/gradbench/blob/aac64c2e32a2baf69a4af49fa98fd3fa0156c016/crates/gradbench/src/main.rs#L965-L968

I'd kind of rather make this into a flag; so, maybe, for each of the variants of the `BadOutcome` enum, we add a flag which causes that variant to just give an exit code of zero. For instance, `--allow-interrupt`, `--allow-timeout`, `--allow-invalid`, and `--allow-error`. Does this sound like a good idea, or is it unnecessary?

samestep:
Oh also: it looks like the build is failing on Windows since [that whole crate is flagged with `#![cfg(unix)]`](https://github.com/jcreekmore/timeout-readwrite-rs/blob/c73d9dfd77d5da8beabb0396be759c89f2815345/src/lib.rs#L9). Is there a cross-platform way to do it instead?

athas:
>  Could you remind me why we'd want this exception to the protocol?

It is useful when writing tools that analyse log files, because it avoids a special case. It is nice to be able to assume that a log file always consists of *2n* lines. It is not a big deal if we assume that only the last line can be "missing", and that this implies a timeout (but it does seem a bit magical, and means we cannot distinguish it from other error modes/crashes).

> For instance, --allow-interrupt, --allow-timeout, --allow-invalid, and --allow-error. Does this sound like a good idea, or is it unnecessary?

This seems easy to implement, and I can easily imagine runs where I want to treat timeouts as an error. I have a hard time seeing when I would be interested in `--allow-interrupt`, `--allow-invalid`, or `--allow-error`, but if they can all be supported easily, then I suppose they are harmless.

> Is there a cross-platform way to do it instead? 

At the system level, certainly, but this is the only crate I could find. I personally don't find it likely that anyone would want to run `gradbench` on Windows outside of WSL, however.

athas:
Actually, I can think of a case where `--allow-invalid` is useful: when the eval has buggy validation, and does not accept a flag to disable validation. (None of our evals have such a flag currently, but it's something I've been meaning to add.)

samestep:
It could also make sense to use `--allow-invalid` in CI after we implement #139.

samestep (approved):
OK, I've fixed the Windows build in 947d170668b566bfe414f45bd7b1b85f6112ebf0 and removed the protocol modification in e948e44cfaae6d611c9c93bc6a69c5b3c5a89e95. Does this seem like a reasonable point to merge this PR? I'm open to revisiting the $2n$ lines thing in the future if having the exception downstream turns out to be too much of a pain.

athas:
Fine by me.

PR #179, opened 2025-01-07
Modify evals and tools to handle closed pipes gracefully.

athas:
Previously, if the stdin/stdout pipes were closed unexpectedly by the intermediary (usually because of an error somewhere), we'd get loud stack traces and nonzero exit codes. This has two problems:

* It obscures the first error (the one that causes the intermediary to close the pipes) because the followup error is just as verbose.

* It makes timeouts (#178) much more verbose, because they inherently depend on unexpectedly shutting down the tool and/or eval.

I think it is correct design for debuggability that the stderrs of tool/eval are not in any way impeded, but it comes with an obligation not to needlessly litter them. It is very unlikely that the change in this PR will suppress a true bug.

### DIFF
--- a/python/gradbench/evaluation.py
+++ b/python/gradbench/evaluation.py
@@ -84,7 +84,10 @@ class SingleModuleValidatedEvaluation:
         print(flush=True)
         if message["kind"] == "end":
             return
-        response = json.loads(sys.stdin.readline())
+        l = sys.stdin.readline()
+        if l == "":
+            raise EOFError
+        response = json.loads(l)
         if response["id"] != self.id:
             raise ValueError(f"expected message ID {self.id}, got {response['id']}")
         self.id += 1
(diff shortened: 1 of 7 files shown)


python/gradbench/evaluation.py
```diff
@@ -84,7 +84,10 @@ def send(self, message: Any) -> Any:
         print(flush=True)
         if message["kind"] == "end":
             return
-        response = json.loads(sys.stdin.readline())
+        l = sys.stdin.readline()
+        if l == "":
+            raise EOFError
```
samestep, line 89: Huh, I didn't realize `readline` would return an empty string in this situation. Can we add a unit test for this? I've been thinking for a while that we should start adding unit tests for things like this that don't happen during "normal" execution (so they're not currently being checked in CI).

samestep: Yeah we don't need to do it in this PR, but it'd be nice to do at some point; I opened  #180 to note this.

samestep (approved):

PR #181, opened 2025-01-09
ba: quiet EOFError and BrokenPipeError.

athas:


samestep (approved):
Thanks! For context, this is following up on #179 and https://github.com/gradbench/gradbench/pull/178#issuecomment-2580668240.

PR #183, opened 2025-01-15
Replace VLAs with std::vector.

athas:
VLAs are not standardised in C++.


samestep (approved):
Seems reasonable to me!

PR #184, opened 2025-01-15
Implement gmm with Enzyme.

athas:
I figured out why my first attempt didn't work. Instead of reading the rather austere documentation, I thought about how it *should* work (somehow passing in a nonzero seed vector for the cotangents) and looked for a way to do something that looks like that.

### DIFF
--- a/tools/enzyme/.gitignore
+++ b/tools/enzyme/.gitignore
@@ -1,2 +1,3 @@
 *.o
 /run_hello
+/run_gmm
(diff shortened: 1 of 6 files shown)


samestep (approved):
Sweet!

So, if

> somehow passing in a nonzero seed vector for the cotangents

wasn't the answer, what was?

athas:
That *was* the answer, and I was confused about this was never talked about in any of the Enzyme documentation. It focuses too much on magical "it just works" pattern matching without describing the underlying mechanisms, so I eventually just assumed that the underlying mechanisms would be sensible.

PR #188, opened 2025-01-17
Make Enzyme-GMM work with multiple runs.

athas:
It may be a good idea to use --runs 2 in CI to ensure these errors don't occur.


samestep (approved):
Thanks!

Regarding using `--runs 2` in CI: sounds fairly reasonable to me, although how would we then automatically check that tools are doing the right thing here? Or do you just mean to catch cases where the tool actually fails or throws an error when asked to do multiple `--runs`?

athas:
Yeah, it's just to catch the most obvious errors. You could also check that multiple runtimes are actually returned.

PR #189, opened 2025-01-17
Make LSTM accept pointer instead of vector for state.

athas:
This is consistent with the other parameters (and other objective functions for that matter), and is necessary for Enzyme to easily accept the lstm_objective function.


samestep (approved):
Sorry for the delay on reviewing this! Looks good to me; I'm not a C++ expert so I'll generally defer to you on C++ parts of our codebase.

PR #190, opened 2025-01-17
Add Enzyme-LSTM, Enzyme-BA, and Enzyme-HT.

athas:
Depends on #189.


samestep (approved):
Amazing! Looks great to me, thanks so much for implementing these!

PR #191, opened 2025-01-21
ht: generate unambiguous workload descriptions.

athas:
Previously many of the workloads would have the same description.

The new descriptions are however so long that they are cut off by the intermediary, but that can be fixed elsewhere.


samestep (approved):
Thanks! This also made me realize there was a bug in how the intermediary truncated the descriptions; I've fixed that here now.

PR #193, opened 2025-01-23
eval proposal: k-means clustering

athas:
This PR contains a new eval: `kmeans`. This is an implementation of *k*-means clustering via Newton's Method applied to an appropriate cost function. The interesting part, from an AD perspective, is that both the Jacobian and Hessian is needed, and the Hessian is sparse - it has nonzero entries only along the diagonal. This eval thus tests a tool's ability to efficiently handle such sparsity, and the efficiency of computing both Hessian and Jacobian.

This PR includes a full implementation of the k-means algorithm, including the Newton's Method convergence loop. Another possibility is to define an eval that includes only the multiplication of the Jacobian by the inverse of the Hessian, since the convergence loop is going to be identical (and irrelevant to AD) in all evals.


samestep:
Superceded by #194.

PR #194, opened 2025-01-23
eval proposal: minimal variant of k-means clustering

athas:
This is a variant of #193 that elides the convergence loop.


athas:
Wow, GitHub's Markdown-math is really bad.

samestep (commented):
Thanks, I like this! A couple questions.

1. Are we sure it makes sense for the clusters $C$ to be the last $k$ elements of $P$? I feel like it'd make more sense for the clusters to be a separate part of the input.
2. Instead of separating the eval into two functions `"cost"` and `"direction"`, could we just have one function, and have the tool return both in a single function call? As far as I'm aware, most autodiff tools compute the function result as part of computing the derivative anyway.

athas:
> Are we sure it makes sense for the clusters C to be the last k elements of P ? I feel like it'd make more sense for the clusters to be a separate part of the input.

It's a common way to initialise these algorithms in practice, but I agree it looks weird when we focus on just a single step. I'll make them a separate input field.

> Instead of separating the eval into two functions "cost" and "direction", could we just have one function, and have the tool return both in a single function call? As far as I'm aware, most autodiff tools compute the function result as part of computing the derivative anyway.

Sure, but then you cannot compute the overhead, which means a tool that is slow in constant terms (but performs AD efficiently) will always look slower than e.g. the C++ tools.

athas:
> I also have another question about this: I thought I remembered you saying that part of the point of this eval is for tools to compute not just the gradient but the Hessian too. However, in the direction function, the tool only returns the gradient, as far as I can tell.

It's a question of semantics whether you call this the "gradient" or the "direction". It is computed by differentiating twice, and it is (the diagonal of) a Hessian matrix.

samestep:
> Sure, but then you cannot compute the overhead, which means a tool that is slow in constant terms (but performs AD efficiently) will always look slower than e.g. the C++ tools.

Fair enough, seems like it'd make sense to leave them separate to allow computing this overhead.

> It's a question of semantics whether you call this the "gradient" or the "direction". It is computed by differentiating twice, and it is (the diagonal of) a Hessian matrix.

I'm not sure I understand, sorry; I had thought that the $k$-means cost function looked like

$$\mathbb{R}^{n \times k} \to \mathbb{R}^{k \times k} \to \mathbb{R}$$

so if we fix the first parameter $P : \mathbb{R}^{n \times k}$ then we have a function $f_P : \mathbb{R}^{k \times k} \to \mathbb{R}$. If we take the gradient then we get

$$\nabla f_P : \mathbb{R}^{k \times k} \to \mathbb{R}^{k \times k}$$

which I thought was the `"output"` to the `"direction"` function. But then where does the Hessian come in?

athas:
The cost function is

```math
f : \mathbb{R}^{k\times d} \rightarrow \mathbb{R}
```

(after fixing points), and then by differentiating with `vjp` (or `grad`) we get

```math
f' : \mathbb{R}^{k\times d} \rightarrow \mathbb{R}^{k\times d}
```

which gives us the gradient. However, to use Newton's method, you also need the second derivative, not just the first. The [Jax implementation](https://github.com/gradbench/gradbench/pull/194/files#diff-9fc0a02becdd3fe80db7ed72949aeaefd768a46b9796ba6b26b0ebdeeb40c832R41-R45) makes it fairly clear.

samestep:
Ah gotcha; sorry, you did already write this in `evals/kmeans/README.md` but for some reason I didn't read it carefully enough.

I guess what I'm asking is: since the Hessian is a diagonal matrix, why not just have the tool return the values along that diagonal? So if we have just one function then it'd have three outputs (`cost`, `gradient`, and `hessian`) or if we have separate functions then the cost, gradient, and Hessian could each be their own function.

athas:
Some tools may be able to compute the product of the Jacobian and the inverse Hessian in a particularly efficient way, without manifesting them both to memory first, so it is useful to have a single function that does both, so they can do that.

samestep:
OK! That all makes sense to me. So, other than making the clusters a separate input field, I'm happy with this PR.

athas:
Good, then I'll also implement it with some more tools and figure out why Jax is being numerically weird.

samestep:
My very naive guess for JAX is that it's because it uses 32-bit floating-point by default instead of 64-bit, but I could be totally wrong.

athas:
It was indeed using single precision. Apparently you have to ask Jax in multiple ways before it will begrudgingly use double precision.

athas:
I had intended to also implement this with Finite, but I think it'll be a little while before I get around to it. Five tools (PyTorch, Jax, Futhark, Manual, and Enzyme) is a pretty decent start, though. I think it will also be easy enough to implement with TensorFlow and SciLean.

samestep (approved):
Looks great to me, thanks so much!

PR #195, opened 2025-01-24
Move ba data into 'evals' subdir instead of 'python'.

athas:
This is consistent with the other evals, and is also more sensible: a bunch of data files is not Python code.


samestep (approved):
Awesome, thanks! I've been meaning to do this for a while, but kept forgetting.

Issue #141, opened 2024-12-02
Validation is too slow


athas:
Gradbench supports validation, but currently all the validation is for the eval to run the PyTorch implementation and compare the result. This has the following downsides:

1. It is inherently inefficient because when running *N* tools with some eval, you also end up running the PyTorch implementation *N* times.
2. The PyTorch implementations are somewhat slow, which makes (1) a significant obstacle for larger workloads.

The basic idea of executing a golden implementation on demand has the problem that it essentially has to run in the same Docker image as the eval. This makes it annoying to simply swap out the PyTorch implementation with some other one, such as the Manual ones from ADBench, as these come with lots of C++ baggage.

This could be addressed by extending the protocol  to allow the eval to communicate with *two* tools (one of which is the "golden" one), but that would be rather complex, so let's not do that. Instead, it should be possible to pass an eval a JSON file that is the result of a previous execution, and contains the golden results, and use that for validation. That JSON file can be computed using whatever tool we trust (hopefully a fast one). I think that was brought up in the initial discussion of the design for validation as well.

samestep:
Personally I'm not a huge fan of the added complexity of having the eval communicate with multiple tools, and I'd be more inclined to just go the route of using the Manual implementations from ADBench inside of the eval itself. I know you mentioned that it's a bit annoying to do this, but it's not too bad, right? Especially since we're going to include Manual under `tools` anyway, per #136.

athas:
We might not have Manual for future evals. In fact, we almost certainly will not. AD is mostly useful for those problems that are completely intractable to differentiate by hand.

samestep:
Yep that is fair. Still though, in general can't we just always use some fast tool for validation in any eval? That means the slowdown is at most 2x. Or we could even just save the expected outputs when we build the Docker image itself, if we really want to save time. All these seem better to me than adding extra complexity to the protocol itself.

athas:
I would prefer to pass in the expected outputs as a command line argument to the eval (which would then have to be copied into the Docker image, but I think that is not so difficult). The specifics can be worked out. I definitely agree that this should be done without modifying the protocol.

athas:
[...]

PR #196, opened 2025-01-24
Validate according to a golden log

athas:
This is a draft of how validation could be done based on a log from running a "golden" tool. It is inspired by the discussion in #141.

The way it works is that the intermediary is provided a golden log file, and whenever the eval receives a response from the tool, that response has also been augmented with a `golden` field that contains the golden response. The eval can use that to perform validation as necessary.

The implementation here is somewhat crude, and I only made the necessary changes for `gmm`.

One change, that I think should be done irrespective of whether this approach to validation is desirable, is that the intermediary actually parses the JSON responses sent by the tool/eval. This makes for much better error messages in case one of them misbehaves.

I also had to extent validation with the notion of "unknown", which happens when there is no golden log file.


athas:
This is pretty much operational, although CI also needs modification: a job is needed to produce a golden log file, which is then used by dependent jobs.

samestep (commented):
Apologies, it's been a little while since I've thought about #141 so I'm paging context back in. On the surface though, I'm not sure I'm in favor of this; seems like it requires modifications to both the protocol and the CLI interface?

Assuming that the reason for this is just to make `gradbench run` faster, is there a reason we can't just do what I proposed in https://github.com/gradbench/gradbench/issues/141#issuecomment-2515130706?

> Or we could even just save the expected outputs when we build the Docker image itself, if we really want to save time.

athas:
If this is not an acceptable approach, then I will need the ability to disable validation in the evals and perform my own offline validation on the log files. That is not so difficult I suppose.

My usage plans for Gradbench makes online validation through recomputation quite impractical (PyTorch is way too slow, and making anything fast enough run as part of the evals is annoying, engineering-wise).

athas:
Example: I will be doing experiments where only optimised GPU execution is fast enough. Modifying the Dockerfiles for the tool(s) to support GPU passthrough is already somewhat annoying (and I may just decide to run those tools outside of Docker), and doing the same thing for the evals just to run a golden GPU tool (which may also involve arbitrary amounts of tool-specific GPU configuration, which must be passed to it via command line options to the eval, or magic environment variables) is just too much trouble.

samestep:
Perhaps I'm misunderstanding. Can you explain why the approach of precomputing the golden results and baking them into the eval's Docker image doesn't work in your use case?

samestep:
Ah I see, so the idea is that it's prohibitive to actually run the reference tool while building the Docker image for the eval? That makes sense. In that case, my suggestion would be this:

1. Configure the eval to do no validation.
2. Run the tool on the eval to produce a log.
3. Commit that log (or some processed version of it) to the repository.
4. Read those precomputed results in the eval's `Dockerfile`.
5. Configure the eval to validate using the precomputed results.

Does this seem reasonable? Or does it still not cover your use case?

athas:
For my use case, I don't actually care what goes on in Gradbench's own CI. Online validation by recomputation is *possibly* fine for any workloads that we are going to bother running on GitHub's infrastructure. Using a golden log file would also work for CI however, and while I was originally skeptical, I have come to like the conceptual elegance of the idea. (I believe you were the first to bring it up when we first discussed validation.)

I do not think committing a golden log file to the repository is a good idea. It's too annoying when you change an eval (which admittedly might not happen very often), and some of the log files can be pretty large (e.g. ba's is 49MiB, although that is uncompressed, although also with only the tiniest datasets enabled).

The reason why I did a protocol extension instead of putting the golden log in the Dockerfile is to avoid having to do anything Docker-specific.

samestep:
Hmm, I think I understand. So in that case, why do you care about making this work with Docker at all? The intermediary already doesn't care whether or not the eval is using Docker, e.g. you can just run the eval directly outside Docker like this:

```sh
./gradbench run --eval 'uv run python/gradbench/gradbench/evals/hello/run.py' --tool './gradbench repo tool pytorch'
```

So, why not just add a flag to the eval to let it take a golden log file as a CLI argument? It wouldn't really work with Docker, but why do you need to run the eval in Docker for your use case anyway?

athas:
I might not need to run the evals in Docker (many HPC systems don't even allow that, sadly), but supporting two different ways of doing validation seems like a bad idea, and we *definitely* want to use Docker in CI (and many other cases).

samestep:
Oh; then I misunderstood. So you're saying, you want to remove the status quo in which each tool comes with its own validator built in, and instead only do validation when the user explicitly specifies `--golden` on the command line? I don't think I'm in favor of that either.

athas:
Fair enough.

PR #197, opened 2025-01-24
Rework DiffSharp tool to not be a script.

athas:
I think this results in faster runtimes, and it also makes it easier to run the tool outside of Docker.


samestep (changes requested):
Sweet! You mentioned running DiffSharp outside of Docker; could you modify `tools/diffsharp/README.md` to say exactly how to do that?

athas:
Done. I had originally planned to also implement the various evals, but it turns out that DiffSharp changed significantly since the version that was used in ADBench, and the new API is a not-particularly-interesting PyTorch clone, so I'll probably not get around to it anytime soon.

samestep (approved):
Makes sense. Thanks for this regardless!

PR #198, opened 2025-01-25
Implement LSTM, GMM, and Hello with CppAD.

### LINKED ISSUE #88: CppAd
Brad Bell emailed me during [AD 2024](https://www.autodiff.org/ad24/) with these links:

> AD functions
> https://cppad.readthedocs.io/latest/base2ad.html
> 
> Atomic functions
> https://cppad.readthedocs.io/latest/atomic_four_define.html#atomic-four-define-title

It'd be nice to include this tool in GradBench.

athas:
Resolves #88.

### DIFF
--- a/cpp/adbench/shared/matrix.h
+++ b/cpp/adbench/shared/matrix.h
@@ -84,12 +84,7 @@ T arr_max(int n, const T* const x)
     T m = x[0];
     for (int i = 1; i < n; i++)
     {
-#ifdef TOOL_ADEPT
-        if (m < x[i])
-            m = x[i];
-#else
-        m = fmax(m, x[i]);
-#endif
+        m = std::max(m, x[i]);
     }
     return m;
 }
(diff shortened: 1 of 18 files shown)


athas:
I figured out how to precompute the tape in the `prepare` step (it was some C++ copy constructor bullshit that made it segfault, so now I just construct it once, with dynamic allocation).

CppAD is a good bit slower than ADOL-C on this one, so perhaps I am doing something wrong.

samestep (approved):
Sweet, thanks for doing this! Not sure what to expect for performance; perhaps @bradbell can weigh in?

athas:
Performance improved a lot when I found the `optimize` method. That doesn't mean it cannot be made better still.

bradbell:
I was working on a similar tool; see
https://cmpad.readthedocs.io/

I just started with gradients but was planning to expand it.

PR #200, opened 2025-01-29
Add commands to both build and run Docker images

### LINKED ISSUE #187: Need to remember to build Docker images before running them
When working in the GradBench repo, it's common to run an eval and tool together after making a small change to one or the other. Currently, this requires the developer to remember to first run `./gradbench repo build-eval` and/or `./gradbench repo build-tool`, before running `./gradbench eval` and `./gradbench tool` as part of a `./gradbench run` command.

It'd be nice for the CLI to automatically check if a Docker image is out of date before running it, and if it is, automatically build it first. Maybe we could call these commands `./gradbench repo eval` and `./gradbench repo tool`, to distinguish them from `./gradbench eval` and `./gradbench tool` which should still work outside the repo.

samestep:
Resolves #187.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,39 +66,27 @@ This script will always automatically rebuild the CLI if it is not already up to
 
 ## Docker
 
-Use the `repo build-eval` subcommand to build the Docker image for any evaluation script:
+Use the `run` subcommand to run a given evaluation on a given tool. You can use pass any commands for the evaluation and tool, but to use the Docker images, the easiest way is to use the `repo eval` and `repo tool` subcommands:
 
 ```sh
-./gradbench repo build-eval hello
-```
-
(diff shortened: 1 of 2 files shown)


athas (approved):

athas:
There is now an unfortunate nomenclature confusion in the documentation: "an evaluation" is now both a thing that is done via the protocol, and the things that live in the `evals` directory. It can be confusing because "running an evaluation" can reasonably be interpreted to refer to either of these concepts.

samestep:
Hmm yeah true; any suggestions for an alternate word?

athas:
I think "eval" is fine, if one can resist the temptation to spell out "evaluation". Simply "benchmark" is the most conventional term.

samestep:
Gotcha; should I go through and replace all our existing usages of the spelled-out word "evaluation" with just "eval"?

athas:
I think that would be the easiest solution.

PR #202, opened 2025-01-29
Simplify C++ infrastructure.

### LINKED ISSUE #199: The C++ API needs a slight refactoring
There's some design features that made sense for ADBench, but not for GradBench. In particular, ADBench uses the same object instance for both objective and gradient computation, which means the `prepare` method often does expensive work in expectation of the latter. This is wasteful in GradBench, where a new instance is created for every evaluation. I kept the code as ADBench-like as possible to ease porting, but I think there is not much left to port.

We also don't use the internal timing loops, so they should be removed.

It is not too hard to do this, but I don't want to do it while there are outstanding PRs that still use the current API.

athas:
Resolves #199. This looks like a large change, but it is actually not significant. It simplifies the IData class to remove some ADBench-isms that are unnecessary in Gradbench. The overall impact is less boilerplate in tool implementations. In particular:

* calculate_objective/jacobian no longer accept a 'times' parameter - this is done in the main function instead.

* The 'prepare' method has been replaced by a constructor.

* A 'prepare_jacobian' method has been introduced for those tools where it makes sense to measure the work done before computing the jacobian.

### DIFF
--- a/cpp/adbench/main.h
+++ b/cpp/adbench/main.h
@@ -24,41 +24,43 @@ int generic_main(int argc, char* argv[]) {
 
   ReadInput(input_file, input, &runs);
 
-  Benchmark b;
-
-  double prepare_time_taken;
-  {
-    struct timespec start, finish;
-    clock_gettime( CLOCK_REALTIME, &start);
-    b.prepare(std::move(input));
-    clock_gettime( CLOCK_REALTIME, &finish);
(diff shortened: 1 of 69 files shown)


samestep (approved):
Looks good to me! I didn't review all the changes in detail so thanks for summarizing them in the PR description. Feel free to merge once CI passes.

Issue #170, opened 2024-12-16
Protocol does not have a way to signal evaluation failure


athas:
As part of debugging a segmentation fault (AD for C++ is a great experience), I realised that there is no way for a tool to reply to an `"evaluate"` message with failure information. You can certainly produce an invalid output and then dump error information to stderr (or even include extra fields and fish them out of the log), but it may be handy to have an explicit way of stating that evaluation failed in a way that was detected, which I feel is different than silently producing incorrect results.

PR #204, opened 2025-01-30
Deduplicate Python modules for C++ tools.

athas:
The convention in Gradbench is that each eval for a tool is represented by a Python module. This unfortunately means that the C++-based tools have dozens of identical modules. This PR addresses this by creating a single module that is then symlinked.

I also made the tools all run the compiler upon receiving 'define' - previously some of them would do so in the Dockerfile itself. I am not convinced this is actually the best approach (due to #170), but I remember Sam preferring it this way.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -66,7 +66,7 @@ This script will always automatically rebuild the CLI if it is not already up to
 
 ## Docker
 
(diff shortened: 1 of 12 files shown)


athas:
This also fixes the currently-broken ADOL-C tool (and #213 ensures the breakage will not repeat).

samestep (commented):
I'm not a huge fan of using symlinks for this. What are your thoughts on an approach like #214 instead?

PR #207, opened 2025-01-30
Speed up Enzyme a bit.

athas:
This is by adding `__restrict__` to the GMM objective function (significant impact, may also help other tools) and telling the compiler not to worry about errno for math functions (much smaller impact).

Enzyme is still inexplicably slow for GMM.


samestep (approved):
Cool. I should really work on #139 to be able to see some indicator of performance improvement when reviewing PRs like this.

PR #212, opened 2025-01-30
Kill entire eval and tool process groups on Ctrl-C

samestep:
Doing this before #178 to figure out the signal stuff now so that that PR can just focus on the actual `--timeout` feature.


athas (approved):
The rules for what you can do in a signal handler are arcane (see `man 7 signal-safety`), and I really have no idea how they translate to Rust. Killing everything around you is probably OK. If this works in practice, then I'm OK, even if it happens to violate some rule.

samestep:
Hmm... looking at [that page](https://man7.org/linux/man-pages/man7/signal-safety.7.html), I see that `kill(2)` is listed as "async-signal-safe" but I don't see `getpgid` or `killpg`. The Google AI Overview says it's fine though? 😛

PR #213, opened 2025-01-31
Detect define and evaluation errors

### LINKED ISSUE #170: Protocol does not have a way to signal evaluation failure

athas:
Resolves #170.

This PR extends the protocol definition and evaluation responses with an "error" field that a tool can use to provide information on what went wrong. This provides two important features:

* It allows us to distinguish a tool missing an eval from compilation of an eval failing.

* It allows evaluation failure to be reported as something beyond an invalid result.

Tools must be modified to actually catch exceptions (or whatever) and stuff them into the responses). This is why this PR is marked as draft: until #204 is merged, this means modifying a lot of otherwise-identical modules.


athas:
I have not modified all the tools yet, but at least the C++-related ones should produce nicely reported errors now (and let's face it, the C++ tools are the ones that will fail the most).

athas:
And to clarify: I don't intend to modify any more tools in this PR.

crates/gradbench/src/main.rs
```diff
@@ -627,29 +635,43 @@ impl<I: Write, O: BufRead, C: FnMut() -> Duration, T: Write, L: Write> Intermedi
                     write!(self.out, " {}", nanostring(nanos).dimmed())?;
                     let response: DefineResponse = self.parse_response(&tool_line)?;
                     self.print_status(response.success)?;
+                    if let Some(error) = response.error {
+                        print!("\n{}", error.red());
```
samestep, line 639:
```suggestion
                        write!(self.out, "\n{}", error.red());
```

crates/gradbench/src/main.rs
```diff
@@ -627,29 +635,43 @@ impl<I: Write, O: BufRead, C: FnMut() -> Duration, T: Write, L: Write> Intermedi
...
-                    let mut first = true;
-                    for (name, (num, ns)) in timings {
-                        if first {
-                            write!(self.out, " {}", "~".dimmed())?;
-                        } else {
-                            write!(self.out, ",")?;
+                    match response.error {
+                        Some(error) => {
+                            self.print_status(false)?;
+                            print!("\n{}", error.red());
```
samestep, line 650:
```suggestion
                            write!(self.out, "\n{}", error.red());
```

README.md
```diff
@@ -145,11 +145,13 @@ type Message = StartMessage | DefineMessage | EvaluateMessage | AnalysisMessage;
 
 interface DefineResponse extends Base {
   success: boolean;
+  error?: string;
```
samestep, line 148: Can you name this `message` instead, for consistency with `"analysis"` messages?

```suggestion
  message?: string;
```

Or, alternatively, if you'd prefer to name this `error`, could you also rename the `message` field in `"analysis"` messages from `"message"` to `"error"`?

README.md
```diff
@@ -145,11 +145,13 @@ type Message = StartMessage | DefineMessage | EvaluateMessage | AnalysisMessage;
...
 }
 
 interface EvaluateResponse extends Base {
   output: any;
   timings?: Timing[];
+  error?: string;
```
samestep, line 154: Same request here; I think also, `output` should be made optional, and a mandatory `success` field should be added.

samestep (changes requested):
Thanks, this has been on the backlog too long! A few requests.

README.md
athas: I will rename them all to `"error"`; on average this is more evocative of what is actually going on.

README.md
athas: Fiddling with `output` belongs in a different PR.

samestep: Hm? I'm not sure I understand; are you saying that in this PR, even if a tool returns an `error`, it must _also_ return an `output`?

athas: That does seem weird, yes. At the Python level it is a `None` value, but I don't actually know how that is serialised to JSON.

If we make it optional, what do we do about an `EvaluateResponse` that has neither `output` nor `error`?

samestep: Neither `output` nor `error` is fine, as long as `"success"` is `false`. If `"success": true` but no `output` then that's a protocol error.

athas: Alright, I'll see if I can do that. This really would be better if Python could express sum types.

samestep: I think Python can express sum types more or else fine; this is really just a matter of whether we want to include that sum type in the protocol or not. I think it could be reasonable to do so.

athas:
All requests have been addressed. Can this be merged soon? Debugging new implementations is very tedious when errors are silently swallowed.

samestep:
Yes, sorry! I'm really bad at reviewing things in a timely manner. Right now I'm adding some unit tests to this PR, and then it should be good to merge. As an aside, CI on this PR is currently failing due to an unrelated issue: #238

samestep (commented):
OK, I've added some unit tests; but now that I've updated the protocol per our discussion in https://github.com/gradbench/gradbench/pull/213#discussion_r1941916178, I'm guessing every tool is going to fail the `run` job by not giving the `"success"` field in response to `"evaluate"` messages.

samestep (approved):
OK @athas I think I've fixed all the tools to use the new protocol. Can you look over this to see if it still looks good to you? If so then feel free to merge once CI passes.

samestep:
And by "once CI passes" I don't include the Tapenade multi-platform build, since that's unrelated.

Issue #180, opened 2025-01-09
Unit testing


samestep:
Some recent improvements to the GradBench CLI have added things that aren't tested in the "normal" flow of CI, such as optional flags or error handling. Examples:

- #155
- #162
- #178
- #179

We should write unit tests for these sorts of things, to make them less likely to silently break in the future.

samestep:
Closing this for now since #226 covers a big chunk of this; can open followup issues on more specific things if we need to.

PR #218, opened 2025-02-03
Validate JSON in intermediary before forwarding

### LINKED ISSUE #215: The intermediary should systematically check that evals and tools produces valid JSON
The intermediary will in most cases parse the JSON lines sent from the tool and eval. However, if either happens to produce invalid JSON, then the error message is not particularly helpful - and worse, tool messages are sent to the eval before they are parsed, which can cause the eval to crash with a confusing error message.

I think the intermediary should at least verify that all messages are valid JSON before they are relayed, and print a reasonable error if they are not ("tool's response to message XX is invalid JSON").

samestep:
Resolves #215.


athas (approved):

samestep:
Marking as draft again until the error messages are improved, before which I think I need to work on #180.

PR #223, opened 2025-02-04
Accept comparisons of floats and ints.

### LINKED ISSUE #222: support for `0` instead of `0.0` in the tool output
Issue: `0` for zero float instead of `0.0` is not parse properly.

Lean generates `0` instead of `0.0` for zero floating point number when producing Json. This causes parsing it as int instead of float
```
[60] eval  gmm::jacobian   32_25_1000            135ms ~        53ms evaluate ✗
Found 1 mismatches, showing 1:
[479]: expected value of type <class 'float'>, got value of type <class 'int'>
 [62] eval  gmm::objective  32_50_1000            114ms ~        23ms evaluate ✓
 [64] eval  gmm::jacobian   32_50_1000            225ms ~       114ms evaluate ✗
Found 32 mismatches, showing 10:
[242]: expected value of type <class 'float'>, got value of type <class 'int'>
[243]: expected value of type <class 'float'>, got value of type <class 'int'>
[244]: expected value of type <class 'float'>, got value of type <class 'int'>
[245]: expected value of type <class 'float'>, got value of type <class 'int'>
[246]: expected value of type <class 'float'>, got value of type <class 'int'>
[247]: expected value of type <class 'float'>, got value of type <class 'int'>
[248]: expected value of type <class 'float'>, got value of type <class 'int'>
[249]: expected value of type <class 'float'>, got value of type <class 'int'>
[250]: expected value of type <class 'float'>, got value of type <class 'int'>
[251]: expected value of type <class 'float'>, got value of type <class 'int'>
```

Example of problematic response of the tool:
```
{ "elapsed": { "nanoseconds": 4401014653 }, "response": {"timings":[{"nanoseconds":110762248,"name":"evaluate"}],"output":[-6.496277,-21.297273, ... ,76.734722,46.864409,-15.393058,-4.644169,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,-163.665279,-173.403034, ... ] }
```

athas:
Fixes #222.

### DIFF
--- a/python/gradbench/gradbench/comparison.py
+++ b/python/gradbench/gradbench/comparison.py
@@ -26,6 +26,12 @@ def compare_json_objects(expected, actual, tolerance=1e-4, path=""):
     """
     mismatches = []
 
+    # Fudge the types a bit.
+    if type(expected) is int and type(actual) is float:
+        expected = float(expected)
+    if type(expected) is float and type(actual) is int:
+        actual = float(actual)
+
     if type(expected) != type(actual):
         mismatches.append(
             f"{path}: expected value of type {type(expected)}, got value of type {type(actual)}"
--- a/python/gradbench/gradbench/eval.py
+++ b/python/gradbench/gradbench/eval.py
@@ -87,7 +87,7 @@ class SingleModuleValidatedEval:
         l = sys.stdin.readline()
         if l == "":
             raise EOFError
-        response = json.loads(l, parse_int=float)
+        response = json.loads(l)
         if response["id"] != self.id:
             raise ValueError(f"expected message ID {self.id}, got {response['id']}")
         self.id += 1
--- a/python/gradbench/gradbench/eval.py
+++ b/python/gradbench/gradbench/eval.py
@@ -87,7 +87,7 @@ class SingleModuleValidatedEval:
         l = sys.stdin.readline()
         if l == "":
             raise EOFError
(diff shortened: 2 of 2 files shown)


python/gradbench/gradbench/comparison.py
```diff
@@ -26,6 +26,12 @@ def compare_json_objects(expected, actual, tolerance=1e-4, path=""):
     """
     mismatches = []
 
+    # Fudge the types a bit.
+    if type(expected) is int and type(actual) is float:
+        expected = float(expected)
+    if type(expected) is float and type(actual) is int:
+        actual = float(actual)
```
samestep, line 33: The second seems fine: if `expected` is a `float` and `actual` is an `int`, it makes sense to coerce. But the first one doesn't seem as fine: if `expected` is an `int` and `actual` is a `float`, that seems like a mistake.

athas: The inputs to this function is currently always JSON objects, so the difference is pretty much just down to arbitrary implementation decisions in (de)serialisers.

samestep: Not quite, right? Doesn't the `expected` come directly from some Python implementation running in-process, with no JSON (de)serialization?

PR #230, opened 2025-02-05
Make it possible to disable validation.

athas:
For simplicity, disabled validation is transmitted as valid at the protocol level. A future extension could make it explicit that the result is not validated.


samestep (approved):
@athas I wasn't a huge fan of making the `validator` be `Optional` in `SingleModuleValidatedEval`, so I removed that and replaced it with just an `approve` validator that evals can use instead. Otherwise looks good to me!

PR #231, opened 2025-02-05
Add "eval", "tool", and "config" fields to start message/response

athas:
This PR adds a mandatory "eval" field to the "start" message, which must contain the name of the eval.

It also adds a mandatory "tool" field to the tool response, which must contain the name of the tool.

Finally, both the message and response may optionally contain a "config" field that can contain arbitrary eval/tool-specific configuration information.

This information is highly useful when writing scripts that process log files. It means you need no information beyond what is in the file itself. It is also very easy to implement, so imposes no real burden on tool authors.


README.md
```diff
@@ -123,6 +123,8 @@ interface Timing extends Duration {
 
 interface StartMessage extends Base {
   kind: "start";
+  eval: string;
```
samestep, line 126:
```suggestion
  eval?: string;
```

README.md
```diff
@@ -147,6 +149,11 @@ interface AnalysisMessage extends Base {
 
 type Message = StartMessage | DefineMessage | EvaluateMessage | AnalysisMessage;
 
+interface StartResponse extends Base {
+  tool: string;
```
samestep, line 153:
```suggestion
  tool?: string;
```

README.md
```diff
@@ -88,7 +88,7 @@ As shown by this example, the intermediary forwards every message from the eval
 
 The session proceeds over a series of _rounds_, driven by the eval. In each round, the eval sends a _message_ with a unique `"id"`, and the tool sends a _response_ with that same `"id"`. The message also includes a `"kind"`, which has four possibilities:
 
-1. `"kind": "start"` - the eval always sends this message first, waiting for the tool's response to ensure that it is ready to receive further messages.
+1. `"kind": "start"` - the eval always sends this message first, waiting for the tool's response to ensure that it is ready to receive further messages. This message and the tool response also contains the name of the eval and tool, and optionally a `"config"` field that contains arbitrary information about how the tool or eval has been configured. This information can be used by programs that do offline processing of log files, but is not otherwise significant to the protocol.
```
samestep, line 91:
```suggestion
1. `"kind": "start"` - the eval always sends this message first, waiting for the tool's response to ensure that it is ready to receive further messages. This message may optionally contain the `"eval"` name, and the response may optionally contain the `"tool"` name and/or a `"config"` field that contains arbitrary information about how the tool or eval has been configured. This information can be used by programs that do offline processing of log files, but is not otherwise significant to the protocol.
```

crates/gradbench/src/main.rs
```diff
@@ -317,6 +317,8 @@ enum Message {
     Start {
         /// The message ID.
         id: Id,
+        /// The eval name.
+        eval: String,
```
samestep, line 321:
```suggestion
```

crates/gradbench/src/main.rs
```diff
@@ -367,6 +369,8 @@ enum Message {
 struct StartResponse {
     /// The message ID.
     id: Id,
+    /// The tool name.
+    tool: String,
```
samestep, line 373:
```suggestion
```

samestep (changes requested):
Thanks! I'm totally fine with adding `"eval"` and `"tool"` fields to the `"start"` messages/responses of all our evals and tools in this repo, and with reserving those names and documenting their types in the protocol as described in `README.md`, but I'd rather not make them mandatory; typically, protocols with an initialization message like this make metadata fields like these optional. Does this seem reasonable?

athas:
It is fine with me to make these optional, but in practice all the support and analysis scripts I write will require these, and all evals and tools will provide them, so I think it'll be a fake flexibility. In general I am highly distrustful of protocols with optional fields.

samestep:
That's fair, and I'm open to revisiting the optionality of it later if it turns out to be that important; perhaps I'll change my mins after reviewing more of your scripts from #169, for instance.

athas:
It is also a shame that the intermediary can no longer print the names of the eval and tool, as I found that a useful reminder of what the results correspond to when scrolling through a terminal log.

samestep:
It still can, right? You just have to use `Option<String>` instead of `String`, yeah?

samestep:
I'll add that real quick.

python/gradbench/gradbench/eval.py
```diff
@@ -73,11 +75,12 @@ class SingleModuleValidatedEval:
     validator: Validator
     id: int
 
-    def __init__(self, *, module: str, validator: Validator):
+    def __init__(self, *, module: str, validator: Validator, config={}):
```
samestep, line 78: Sorry, I missed this before; I'd also rather this go in the `start` method instead.

samestep (changes requested):

samestep (approved):

Issue #227, opened 2025-02-04
QEMU is slow


samestep:
Currently we use [`docker build --platform`](https://docs.docker.com/build/building/multi-platform/#build-multi-platform-images) to build cross-platform Docker images in nightly CI:

https://github.com/gradbench/gradbench/blob/2ee294ed6057f87919a68007e6145d16fc8d29c2/crates/gradbench/src/main.rs#L261

https://github.com/gradbench/gradbench/blob/2ee294ed6057f87919a68007e6145d16fc8d29c2/crates/gradbench/src/main.rs#L286

https://github.com/gradbench/gradbench/blob/2ee294ed6057f87919a68007e6145d16fc8d29c2/.github/workflows/nightly.yml#L68

https://github.com/gradbench/gradbench/blob/2ee294ed6057f87919a68007e6145d16fc8d29c2/.github/workflows/nightly.yml#L109

We achieve this by first setting up QEMU:

https://github.com/gradbench/gradbench/blob/2ee294ed6057f87919a68007e6145d16fc8d29c2/.github/actions/docker/action.yml#L13-L15

This is very slow. It'd be nice if we could instead build native images on separate x86 and ARM runners, and then combine those into multi-platform Docker images.

PR #233, opened 2025-02-09
Add Floretta tool

samestep:
This PR has a few little neat things:

1. This initial GradBench implementation for the tool doesn't report any `timings` in response to `evaluate` messages, catching a bug in `python/gradbench/gradbench/eval.py` which I've now fixed.
2. Some of the commands in the Dockerfile need to specify the [Rust target](https://doc.rust-lang.org/rustc/platform-support.html), which depends on [`TARGETARCH`](https://docs.docker.com/reference/dockerfile/#automatic-platform-args-in-the-global-scope); I map between Docker and Rust targets via a small Python script.
3. The Dockerfile uses a [multi-stage build](https://docs.docker.com/build/building/multi-stage/) to produce an image smaller than 20 megabytes, which is one or two orders of magnitude smaller than our other Docker images.
4. Using `cargo run` in the `gradbench` shell script previously worked because there was only one Rust binary to be run; after this PR, it instead works because the GradBench CLI is selected via the [`default-run`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-default-run-field) field in `Cargo.toml`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,33 @@
 version = 4
 
 [[package]]
+name = "addr2line"
+version = "0.24.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "ahash"
+version = "0.8.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


samestep:
Yeah my current plan is to write them directly in WebAssembly; if I change my mind then I'll also change the Docker setup for this, since the tool should always work directly from source text when it gets a `define` message.

samestep:
I suspect that this may be blocked until we switch our multi-platform Docker image build strategy as suggested in #227; see also: https://stackoverflow.com/q/79425799/5044950

samestep:
Never mind, I stand corrected.

PR #242, opened 2025-02-13
Check response.success after response.error

athas:
Also increment failure count on define failure.

Without this change, a tool responding with a failure to a 'define' message is counted as undefined, which means gradbench (in its default configuration) will not return a nonzero exit code. This is why kmeans is actually broken for enzyme/manual right now, without the PR having failed.

### DIFF
--- a/crates/gradbench/src/main.rs
+++ b/crates/gradbench/src/main.rs
@@ -1476,7 +1476,7 @@ mod tests {
         colored::control::set_override(false);
         let result = intermediary.run();
         write_goldenfile("define_error.txt", &intermediary.out);
-        assert_eq!(result, Err(BadOutcome::Undefined));
+        assert_eq!(result, Err(BadOutcome::Failure));
     }
 
     #[test]
--- a/crates/gradbench/src/main.rs
+++ b/crates/gradbench/src/main.rs
@@ -696,17 +696,18 @@ impl<
                     self.print_left(WIDTH_DESCRIPTION, "")?;
                     write!(self.out, " {}", nanostring(nanos).dimmed())?;
                     let response: DefineResponse = self.parse_response(&tool_line)?;
-                    if !response.success {
-                        undefined += 1;
-                    }
                     self.print_status(response.success)?;
                     line.end(&mut self.out)?;
                     if let Some(error) = response.error {
+                        failure += 1;
                         writeln!(self.out, "{}", error.red())?;
                         if response.success {
                             return Err(anyhow!("tool reported success but gave an error"));
                         }
+                    } else if !response.success {
+                        undefined += 1;
                     }
+
                 }
                 Message::Evaluate { .. } => {
                     write!(self.out, " {}", nanostring(nanos).dimmed())?;
--- a/crates/gradbench/src/main.rs
+++ b/crates/gradbench/src/main.rs
@@ -707,7 +707,6 @@ impl<
                     } else if !response.success {
                         undefined += 1;
                     }
-
                 }
                 Message::Evaluate { .. } => {
                     write!(self.out, " {}", nanostring(nanos).dimmed())?;


athas:
Note that this changes a behaviour that was explicitly tested for. I think the new behaviour is superior, as otherwise there is still no way to distinguish define-failure from define-undefined.

samestep (changes requested):
I understand the reasoning behind this PR, but I don't think this is the right way to do it. Specifically, the `"success"`/`"error"` pair are supposed to be semantically the same in `"define"` messages and `"evaluate"` messages: `"success"` actually says whether it succeeded or not (and should be the determinant of which `BadOutcome` variant gets used) and `"error"` is optional metadata for debugging etc. So semantically I think it's wrong to determine good vs bad via the presence of `"error"` if `"success": false`.

Of course, the real goal of this PR is to catch in CI when a tool _should_ implement an eval but fails to do so. I think the solution to this is not to change the semantics of the protocol, but instead just to change the way we do the `run` job in CI, especially since this PR would only catch the issue in cases where the tool does happen to report an `"error"` instead of just `"success": false`; see #139 and related past discussions.

Currently, the `gradbench repo matrix` subcommand just returns a list of all the evals and all the tools:

```
eval=["ba","gmm","hello","ht","lstm"]
tool=["adept","adol-c","autograd","cppad","enzyme","finite","floretta","futhark","jax","manual","mygrad","pytorch","scilean","tapenade","tensorflow","zygote"]
```

We use these in the `eval` and `tool` jobs, and then also in the `run` job like this:

https://github.com/gradbench/gradbench/blob/9ce82a49cc015a106c13d6d84dd5a0f1c60147fd/.github/workflows/build.yml#L180-L182

So, expanded out in a particular CI run, we get this:

```yaml
matrix:
  eval: ["ba","gmm","hello","ht","lstm"]
  tool: ["adept","adol-c","autograd","cppad","enzyme","finite","floretta","futhark","jax","manual","mygrad","pytorch","scilean","tapenade","tensorflow","zygote"]
```

Then GitHub Actions interprets this to mean that we want all pairs. But really, we don't: we know which tools currently support which evals, and for those, we want to check that it still works, and for the tool/eval pairs that are unsupported, we don't even want to start up CI jobs for those at all because that increases pressure on the GitHub Actions concurrency limit. So what we really want is for `gradbench repo matrix` to also give either an explicit list of [all pairs we _do_ want to check](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#expanding-or-adding-matrix-configurations), _or_ an explicit list of [all pairs we _don't_ want to check](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#excluding-matrix-configurations); for instance:

```
exclude=[{"eval":"ba","tool":"autograd"},{"eval":"gmm","tool":"autograd"},{"eval":"ht","tool":"autograd"},{"eval":"lstm","tool":"autograd"},{"eval":"ba","tool":"cppad"},{"eval":"ht","tool":"cppad"},{"eval":"ba","tool":"floretta"},{"eval":"gmm","tool":"floretta"},{"eval":"ht","tool":"floretta"},{"eval":"lstm","tool":"floretta"},{"eval":"ba","tool":"jax"},{"eval":"gmm","tool":"jax"},{"eval":"ht","tool":"jax"},{"eval":"lstm","tool":"jax"},{"eval":"ba","tool":"mygrad"},{"eval":"gmm","tool":"mygrad"},{"eval":"ht","tool":"mygrad"},{"eval":"lstm","tool":"mygrad"},{"eval":"ba","tool":"scilean"},{"eval":"gmm","tool":"scilean"},{"eval":"ht","tool":"scilean"},{"eval":"lstm","tool":"scilean"},{"eval":"ht","tool":"tensorflow"},{"eval":"lstm","tool":"tensorflow"},{"eval":"ba","tool":"zygote"},{"eval":"gmm","tool":"zygote"},{"eval":"ht","tool":"zygote"},{"eval":"lstm","tool":"zygote"}]
```

And then in `build.yml` and `nightly.yml`, we'd write this:

```yaml
matrix:
  eval: ${{ fromJSON(needs.matrix.outputs.eval) }}
  tool: ${{ fromJSON(needs.matrix.outputs.tool) }}
  exclude: ${{ fromJSON(needs.matrix.outputs.exclude) }}
```

Or if we were using `include` instead of `exclude` then we'd just write this:

```yaml
matrix:
  include: ${{ fromJSON(needs.matrix.outputs.include) }}
```

Then we'd also remove `BadOutcome::Undefined` as a special case from the intermediary:

https://github.com/gradbench/gradbench/blob/9ce82a49cc015a106c13d6d84dd5a0f1c60147fd/crates/gradbench/src/main.rs#L1040-L1044

Thoughts?

athas:
It is more work to maintain such a list - and easy to accidentally overlook. I like the elegance of the current approach, I just don't like that so many errors become interpreted as an unimplemented eval.

samestep:
I don't quite follow the argument that it's more work to maintain the list; isn't this just standard software testing?

If we don't add something like this, I don't see how we can automatically catch cases where a tool silently stops supporting an eval without reporting an error.

athas:
> I don't quite follow the argument that it's more work to maintain the list; isn't this just standard software testing?

Yes, but that is also vulnerable to false positives, if you forget to add your contribution to the list.

> If we don't add something like this, I don't see how we can automatically catch cases where a tool silently stops supporting an eval without reporting an error.

This is true, so we are essentially left with the question of which error is *least* likely to arise, and picking that as the one we do not detect. I think a reasonable argument can be made that adding an element to a whitelist is the least likely error, as presumably it can be made only in a PR that is specifically adding a new implementation (meaning reviewers are inclined to check for it).

samestep:
I think in this case we can have our cake and eat it too: we could have a separate job that goes through all the excluded pairs, and double-checks that all of then return `BadOutcome::Undefined`. That way, CI would fail if someone forgets to say that their PR added support for a new eval/tool pair, and also the other way around.

athas:
I think the list should be *positive*, in that it should describe the eval/tool combinations to run. This is because it will then only have to be updated when someone actively adds an implementation. In contrast, if we have a list of excluded pairs, then whenever someone adds a new eval, they have to list every tool that does not yet support it (which might be many). Same thing occurs when adding a new tool (although to a lesser degree, as we are unlikely to have that many evals, even when Gradbench becomes old and mighty).

PR #243, opened 2025-02-13
Fix various kmeans problems

athas:


samestep (approved):
Looks good to me, thanks!

PR #245, opened 2025-02-13
Add Nix setup with shell.nix

athas:
This is useful for people who don't want to run everything inside Docker, but still want some measure of a predictable environment.

Also slightly modifies Enzyme to make it run well both within Docker and Nix.

This shell.nix does not pin Nixpkgs, which is strictly speaking not ideal, but pinning would just add an extra maintenance burden. I'll add pinning if/when I regret not doing so in the first place, or when others also start using this shell.nix.


samestep (approved):
Sweet, thanks!

PR #247, opened 2025-02-13
Check each eval/tool pair's outcome in CI

samestep:
Following up on https://github.com/gradbench/gradbench/pull/242#pullrequestreview-2615644719, this PR does the following:
- Make the `gradbench run` subcommand's exit code correspond to the `BadOutcome` variants.
- Add a new `gradbench exit-code` subcommand which gives the numerical exit code corresponding to a given named outcome.
- Modify the `gradbench repo matrix` subcommand to return a `run` matrix listing all the eval/tool pairs and the expected `outcome` of each one.
- Add an `evals.txt` file for each tool which lists all the evals for which that tool is expected to run with `success`; for all other evals, it is expected to be `undefined`.
- Modify the `run` job in CI to check that `gradbench run` returns the expected outcome for each pair.


athas:
Alright, but then you still need my change from #242 (or equivalent), as otherwise failure-to-define is interpreted as undefined.

samestep:
Right, this PR is still a draft; once I observe which pairs fail, I'm going to implement the system I described in https://github.com/gradbench/gradbench/pull/242#pullrequestreview-2615644719. My current plan is to just have `tools/*/exclude.json` be an optional file that tools can have if they don't support all evals.

samestep:
OK @athas this PR is ready for review! Let me know if this seems like a reasonable approach. The only thing I can think of is if in the future we decide we want to allow _multiple_ possible outcomes for a given eval/tool pair, e.g. either `success` or `timeout`. I'm thinking that we could either modify the `run` matrix to allow multiple outcomes and check against the whole list (seems a bit clunky), or just add an `--allow-timeout` flag like we previously discussed in https://github.com/gradbench/gradbench/pull/178#issuecomment-2635041387. Thoughts?

athas (approved):
I cannot decide whether it would be better for each eval to have a list of supported tools, or each tool to have a list of supported evals. I suspect the latter, since we are likely to add more tools than more evals, and therefore that would conflict less. It is unlikely to have a major impact.

Time will tell whether we will have tools that are just on the boundary between timing out and succeeding in CI, and thus become flaky. I hope not.

samestep:
Fair enough, in 10519c0df6a812e0cb8b90041e1ba1b805b09d56 I've switched it from `evals/*/tools.txt` to `tools/*/evals.txt`.

PR #224, opened 2025-02-04
SciLean implementation of GMM eval

lecopivo:
Implementation of GMM eval with SciLean

PR #250, opened 2025-02-14
Mathlib synch script

lecopivo:
Script to update lean-toolchain, lakefile.lean and lake-manifest.json to match the ones in `tools/scilean`

### DIFF
--- a/docker/mathlib4/synch_with_scilean.py
+++ b/docker/mathlib4/synch_with_scilean.py
@@ -1,0 +1,56 @@
+import json
+import shutil
+import subprocess
+from pathlib import Path
+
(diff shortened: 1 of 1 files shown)


samestep:
Thank you! I've used this script to generate #268. First let's make sure that this actually lets us fix the build for #224, then we'll come back and discuss whether this is the right way to update the mathlib image going forward.

samestep:
Marking as draft since CI is failing. Since this doesn't seem to fix the #224 build issues, though, do we still want to merge this? I'm still thinking that it might be better for me to instead switch to native builds as described in #227, so that way we don't need to build a base image like this. Or, alternatively, I'm thinking it may be desirable to require that any base images like this be built outside of this repository. Thoughts?

samestep:
Superceded by #321.

PR #251, opened 2025-02-15
Further simplify C++ infrastructure.

athas:
The goal is for each implementation to consist of two files: a run_foo.cpp file that does all the real logic, and a trivial foo.py file. This is in contrast to the current situation of three nontrivial C++ code files.

It is still possible to override this and do arbitrarily complicated things for the tools that need this.


athas:
There is now very little boilerplate left, except what is necessary due to C++. This PR also removed the hardcoding of the notions of an "objective" and "jacobian" from the C++ infrastructure, so we can easily support evals with arbitrary (and arbitrarily named) functions. This is currently used by `hello` and `kmeans`.

PR #254, opened 2025-02-18
Document squared distance in k-means

samestep:
Followup on #194 since I didn't see any square roots in the existing implementations; @athas is this correct?


athas (approved):
Either works, actually.

PR #256, opened 2025-02-18
Implement `llsq` and `lse` evals with Floretta

### LINKED ISSUE #327: Floretta multi-platform build is slower than it should be
It just needs to build two static binaries:

https://github.com/gradbench/gradbench/blob/bfaba66ca32404d65e6d781935ce8f209fb7d497/tools/floretta/target.py#L5-L8

However, currently it just uses QEMU for the ARM build, which is slow as mentioned in #227: it takes about an hour. If possible, it would be better to not use emulation, and instead just cross-compile both from x86; see this for an example: https://github.com/ipvm-wg/homestar/blob/0f0a4f5b4c00c856c6eb865da087c0332b119e8d/docker/Dockerfile

### LINKED ISSUE #485: Missing: lse-floretta

samestep:
Resolves #327, resolves #485.


athas:
I don't envy you for implementing numerical algorithms in what is essentially assembly language.

samestep:
Eh, it's really not that bad, it's more just like a very verbose Lisp with no macros.

athas:
And no lists.

samestep:
So yeah it's basically a very verbose C I guess?

samestep:
In place of `malloc` I can just use bump allocation so that should be fine I think; but yeah when I get to the others I may have to give up on `*.wat` and switch to another language to compile from.

samestep:
@athas apologies for the confusing branch name, I'll get to `kmeans` later; just requesting your review here because I rearranged some stuff in `shell.nix`.

athas (approved):
I am OK with the shell.nix changes.

PR #264, opened 2025-02-18
Use "manual" as golden tool for all evals except "hello"

athas:


athas:
Addresses #141. This is still relatively slow to validate, because the Python code that compares the (fairly large) JSON objects is slow, but it's a *lot* faster than having to wait for PyTorch as well.

athas:
As an example, with pytorch-based validation, `ht` with `--max 4` takes six and a half minutes, but only 29s with manual-based validation.

samestep (approved):
Seems good to me!

PR #272, opened 2025-02-19
Add Haskell tool

athas:
Only supports 'hello' for now, but maybe I can add the others.


athas:
Performance isn't exactly admirable.

image
image
image

It is very fishy that the Enzyme overhead is about the same.

athas:
This PR fails because the `kmeans` implementation fails for some inputs. However, this is *intentional* - they exceed the (fixed) allocation limit I imposed. This is another one of those "benign errors" that I would like not to result in a CI error.

samestep:
Yep so if we expect the `haskell` tool to give `gradbench exit-code error` on the `kmeans` eval, do you think it'd be reasonable to be able to specify this by listing `kmeans` in an `evals-error.txt` file? Or do you think it'd be a better idea to just change it to `tools/*/evals/json` which is a mapping from eval names to exit-code names?

athas:
It would be most elegant if it could be put in `evals.txt` itself (perhaps by extending it to an `evals.json` format). I'm not a fan of having to piece together many different files.

samestep:
Yep, makes sense. Honestly, I'm considering even using YAML for this... I just don't like not having trailing commas for this list in JSON. But either sounds fine to me.

athas:
Strongly cautioning against the use of YAML here, or anywhere.

athas:
Anyway, I fixed this one by making it use less memory. On further reflection, I think allowing per-workload failures is an unnecessary complexity. It is sufficient if we handle tool crashes (due to OOM) as robustly as we handle timeouts.

samestep (approved):
Looks great to me, thanks for implementing this!

Just to make sure I understand what the plan is, can you make a GitHub issue describing the proposed way to handle the kinds of failures we were talking about in this thread? I got a bit lost in the possible solutions.

PR #273, opened 2025-02-19
OCaml tool

athas:


samestep (commented):
Similar to the code we've ported from ADBench, could you be sure to include comments linking to the original source for any code you've copied from, e.g., Jesse Sigal in this PR?

athas:
Yes I will do that, but this is so embryonic that I don't really know how much of it I will keep and how much I will have to rewrite. I also hope to convince Jesse to use GradBench as the "official" version of his implementation.

athas:
This is now fully operational. Performance is less than the program transformation tools, but better than many of the C++ tape-based tools.

samestep (approved):
Sweet!

athas:
Owl apparently does not support ARM. Can we use the same trick as for SciLean?

athas:
[In the future Owl will support ARM](https://github.com/owlbarn/owl/issues/684), but the future is not now.

samestep (changes requested):
@athas from that thread you linked, it looks like Owl already supports ARM, but there just hasn't been a release since then; for instance, owlbarn/owl#609 seems to suggest that you can use it via `opam pin`. Could you try that for this PR? I'd really rather avoid allowing x86-only tools whenever possible.

athas:
No, I do not know how to do that with the OCaml build system.

samestep:
It doesn't work to run `opam pin` in the `Dockerfile`?

samestep (approved):

PR #281, opened 2025-02-22
Pass -Bj when building manual tool in eval Dockerfiles.

athas:
This is to ensure we get a fresh build, no matter what we might have copied in from the build host.


samestep (commented):
Instead of doing this, do you think it might make more sense to put the built files in `.dockerignore` similar to #267?

athas:
It would require updating those `.dockerignore` files to keep up with build system changes, while passing `-B` will always do what we want.

samestep (approved):
Fair enough!

Issue #216, opened 2025-02-03
README should be more helpful to contributors


samestep:
From @lecopivo [on](https://discord.com/channels/1242864568850579487/1267596255886970952/1336064860860977323) [Discord](https://discord.com/channels/1242864568850579487/1267596255886970952/1336071060927414283):

> when trying to get started with gradbench I was comptetly baffled by the main readme file as it does not say how are you suppoed to use the library. It is some high level specification of the protocol. Reading the CONTRIBUTING.md was way way more helpful to get me started when adding gmm eval for scilean.

> - the first two commands in README.md should just work (i.e. add `./`)
>   - be more explicit that I should be looking at CONTRIBUTING.md rather then README.md if I want to implement more tools and evals
>   - be explicit about the inputs and outpus of evals - give concrete examples with concrete json
>   - ability to run particular tool on particular eval/data set of eval and print out input and output to terminal

samestep:
Relevant talk mentioned by @athas: https://fosdem.org/2025/schedule/event/fosdem-2025-4990-how-to-write-a-killer-readme/

athas:
I just had a group of students who couldn't get GradBench to work via the README instructions. It worked once I told them to work in a checkout of the repository, use the `./gradbench` script, and use the `repo tool`/`repo eval` commands.

One problem is that the command in the README seems to fail since Floretta was added:

```
$ cargo install --locked --git https://github.com/gradbench/gradbench
    Updating git repository `https://github.com/gradbench/gradbench`
error: multiple packages with binaries found: gradbench, gradbench-floretta. When installing a git repository, cargo will always search the entire repo for any Cargo.toml.
Please specify a package, e.g. `cargo install --git https://github.com/gradbench/gradbench gradbench`.
```

The solution is to use `cargo install --locked --git https://github.com/gradbench/gradbench gradbench`.

However, when that is done, many things still fail:

```
$ gradbench run --eval 'gradbench eval ba' --tool 'gradbench tool enzyme'
  [0] start ba (enzyme)
  [1] def   ba                                  9.794 s ✓
  [2] eval  ba::objective   ba1_n49_m777...         7ms ~         1ms evaluate ✓
  [4] eval  ba::jacobian    ba1_n49_m777...     1.400 s ~        12ms evaluate,         0ms prepare ✗
Found 4 mismatches, showing 4:
Unexpected key 'BASparseMat'.
Key 'cols' is missing.
Key 'vals' is missing.
Key 'rows' is missing.
[...]

PR #282, opened 2025-02-22
Replace --runs with --min-runs and --min-seconds

athas:
The problem with --runs is that a fixed number of runs for all workloads is not appropriate. You want many runs for small workloads, and few for the big ones. This PR removes the --runs option from the evals and replace it with two new options, with the following semantics:

  * --min-runs=N: run each workload at least N many times. Defaults to 1, and must be greater than zero.

  * --min-seconds=N: run each workload until the aggregrate measured runtime is at least this many seconds. May be a fractional number. Defaults to 1.

This should not significantly increase time taken in CI (because it will not by default rerun any workload that already takes a second), but it should dramatically improve the reliability of runtime measurements for smaller workloads.


samestep (changes requested):
The other main issue I have with the current incarnation of `--runs` is that it's not documented in `evals/*/README.md`. While you're making changes in this PR, could you also document this new way, specifically the fields it adds to the `"input"` which tools need to handle?

athas:
I did notice that, but I'm already somewhat bothered by the replication across the eval protocol documentation. I would like to have a single point of reference that summarises the things that are common to all evals (except `hello`, which is the odd one out in many respects). Would you prefer that such shared information goes in `README.md` or in `CONTRIBUTING.md`? In the long run, a `PROTOCOL.md` may be an even better option.

athas:
Interestingly, many of these CI failures are due to tool implementations that cannot handle doing more than one run. Ferreting out such cases is another advantage of this change. For simplicity, I'll include fixes in this PR.

README.md
samestep, line 1: Thanks, this seems like a reasonable place to document it! And then when I get around to addressing #216, I'll move this protocol documentation either to the bottom of `CONTRIBUTING.md` or perhaps to a separate `PROTOCOL.md` file as you suggested.

For now, one request: can you document the semantics of the `Runs` interface and its `min_runs` and `min_seconds` fields? All the other types here have written documentation of their semantics in the **Specification** section; for these, since they're not part of the main protocol itself but just used by some evals, you could just write `/** doc comments */` on the `interface` or on the fields themselves.

athas: I have documented it as part of **Specification**, for commonality.

PR #284, opened 2025-02-25
Make intermediary report average runtimes

athas:


samestep (commented):
Hmm... I did intentionally _not_ do this at first, since semantically I was thinking of multiple timings with the same name only being meaningful as a sum; for instance, maybe a given timing `foo` is split into three different parts, and so the tool reports three different `foo` timings, but they're all kind of different and so it doesn't make sense to take an average. But that's all very abstract and the behavior you suggest seems more useful for what we're doing right now, so this seems fine. I'm going to make a slight tweak and then merge.

Issue #244, opened 2025-02-13
The nightly job fails due to GitHub size restrictions


athas:
The log files produced by the nightly jobs are stored in a branch of the `gradbench` repository. This is pretty handy as it means they are readily available from any checkout. The downside is that this means a log file is limited to at most 50 or 100MiB (I forget which), and the recently added `kmeans` tool exceeds that, because the log files contain the input/output of every evaluation.

I think it is undesirable that evals must be subject to GitHub's constraints when it comes to which workloads they run by default. However, the data produced by the nightly jobs is low quality anyway (single runs, unpredictable hardware), so perhaps it is fine if it is intended solely as demo data.

One solution that would significantly shrink the size of the log files is to remove the input/output fields (or replace them with `null` to indicate pruning). The surrounding metadata is pretty small. This is my preferred solution.

Another is to store the log files somewhere else, that allows larger files. It is absolutely imperative that such storage still does some form of deduplication or compression across files and versions, as otherwise it will quickly fill up, as the nightly job probably produces 1GiB of uncompressed data per run, which is likely to increase in the future, and *the vast majority is significantly duplicated input/output data*.

athas:
As a short term fix, I will shrink the default kmeans datasets.

athas:
This is fixed now, but I still think we should prune the input/output fields of log files.

samestep:
@athas currently I'm not in favor of that, but feel free to open a followup issue for it if you'd like.

PR #270, opened 2025-02-18
Generate stats

samestep:
Based off of penrose/penrose#988, this PR resolves #121, resolves #122, resolves #139, and resolves #140 by doing the following:

1. Rename `gradbench repo summarize` to `gradbench repo stats`, which now summarizes performance data instead of just stating which tools support which evals.
2. Write documentation in `evals/*/README.md` specifying that the _number_ of `"timings"` with `"name": "evaluate"` is considered significant.
3. Run `gradbench repo stats` in CI for every commit, not just nightly builds.
4. On non-nightly builds by contributors with write access to this repo, commit stats to a separate branch and generate a comment on the PR linking to that branch.
5. On the website, use the high-level scores (sum of average run times for all workloads) to color-code the squares corresponding to all eval/tool pairs in the table.
6. Use [Vega-Lite](https://vega.github.io/vega-lite/) to draw charts similar to the ones from the `scripts` branch: #169
7. Make the charts interactive, so different tools can be toggled on and off by clicking on the legend.

The website itself is not yet updated because it builds from `main` and not from this PR, so the link in the PR comment below by **github-actions** will not work. You can try it out locally by checking out this PR and running this command, as described in `CONTRIBUTING.md`:

```sh
bun run --filter=gradbench dev --host
```

Then navigate to http://localhost:5173/?commit=6c5d3b769dc6817733cd3a52212314bdef76dec8 to see the charts for this PR. Here is a screen recording showing what the updated website looks like:

https://github.com/user-attachments/assets/864ebd3e-b629-4863-87df-57026d58da04

PR #287, opened 2025-02-26
ht: use all workloads by default

athas:
This also adds a 10 minute timeout to tool responses in CI, because not all tools are fast enough to handle the larger ht workloads in a reasonable time. Hence this PR is also intended to ensure that the timeout logic works.

The PR is however marked as draft because the log files are so large (878MiB) that they would lead to a repeat of #244. We need to come up with a solution to that problem before we can run more workloads in CI.


samestep:
Just as a heads up, the log file size issue will be resolved once we merge #270; sorry for being slow on that one.

athas:
Thanks for that. The timeout here also seems to hit some of the existing tools, but maybe that's actually fine! They are really slow. I'll ponder making a separate PR just for the timeout stuff.

athas:
Those graphs are hilarious!

samestep:
Exactly what I was gonna say!

Issue #286, opened 2025-02-26
Failure when running GMM with Futhark locally


samestep:
On commit 26efa17f68837880356a0a2036db88bf3c7749f6, on my 2020 MacBook Pro with M1 chip:

```
$ ./gradbench run --eval './gradbench repo eval gmm' --tool './gradbench repo tool futhark'
  [0] start gmm (futhark)
  [1] def   gmm                                 2.409 s ✓
  [2] eval  gmm::objective  2_5_1000            2.398 s ~         0ms evaluate × 17507 ✓
  [4] eval  gmm::jacobian   2_5_1000            1.675 s ~         0ms evaluate × 4306 ✓
  [6] eval  gmm::objective  2_10_1000           1.378 s ✗
unsupported operand type(s) for /: 'NoneType' and 'float'
  [7] eval  gmm::jacobian   2_10_1000           1.400 s ~         0ms evaluate × 2374 ✓
  [9] eval  gmm::objective  2_25_1000           1.477 s ~         0ms evaluate × 5815 ✓
 [11] eval  gmm::jacobian   2_25_1000           1.169 s ~         0ms evaluate × 1046 ✓
 [13] eval  gmm::objective  10_5_1000           1.311 s ~         0ms evaluate × 3778 ✓
 [15] eval  gmm::jacobian   10_5_1000           1.124 s ~         1ms evaluate × 724 ✓
 [17] eval  gmm::objective  2_50_1000           1.250 s ~         0ms evaluate × 3086 ✓
 [19] eval  gmm::jacobian   2_50_1000           1.089 s ~         1ms evaluate × 527 ✓
 [21] eval  gmm::objective  10_10_1000          1.169 s ~         0ms evaluate × 1949 ✓
 [23] eval  gmm::jacobian   10_10_1000          1.071 s ~         2ms evaluate × 366 ✓
 [25] eval  gmm::objective  20_5_1000           1.104 s ~         0ms evaluate × 1113 ✓
 [27] eval  gmm::jacobian   20_5_1000           1.067 s ~         3ms evaluate × 315 ✓
 [29] eval  gmm::objective  32_5_1000           1.055 s ~         2ms evaluate × 473 ✓
 [31] eval  gmm::jacobian   32_5_1000           1.047 s ~         6ms evaluate × 148 ✓
 [33] eval  gmm::objective  20_10_1000          1.060 s ~         1ms evaluate × 576 ✓
[...]

PR #288, opened 2025-02-26
futhark: fail noisily when runtime information is missing

athas:
This should make problems like #286 easier to diagnose.


samestep:
I think another important part of this is that currently the Futhark tool doesn't capture the full stack traceback:

https://github.com/gradbench/gradbench/blob/441eea0c2a0d4a9186c716eb931aa83051625100/tools/futhark/run.py#L37-L38

It may be better to use [`traceback.format_exception`](https://docs.python.org/3/library/traceback.html#traceback.format_exception) like we started doing for evals in #113.

Issue #278, opened 2025-02-20
The intermediary should handle tool crashes more gracefully


athas:
If a tool crashes or otherwise misbehaves grossly, the resulting log file will be syntactically invalid. In practice, tool crashes are going to occur in out-of-memory conditions, as they are very difficult to actually handle gracefully (and such handling is specific to each tool - we really cannot assume it is possible). Essentially, a tool crash should be handled equivalently to a tool timeout, which was done in #178.

PR #289, opened 2025-02-26
Faster and cleaner Haskell kmeans

athas:
Unfortunately, it also uses more memory - so much that it crashes. This means #278 must be addressed before this can be merged.


samestep (approved):
Cool! To be honest though, I don't really see much improvement when I look at the charts. But it seems like the code improvement is good anyways.

athas:
That's because the scale is logarithmic, but the difference is often about 3x.

PR #290, opened 2025-02-26
Use ten minute timeout in CI

athas:
This extends the 'evals.txt' file such that a line can contain 'timeout' as the second word. If so, the tool is expected to time out in CI.


samestep:
This seems fine to me, but it conflicts with some of the refactoring I'm doing in #270. Let's wait just a bit; by the end of tomorrow I should be able to merge #270 and then make this change.

athas:
Now the second word can be the string representation of *any* `BadOutcome`. This allows some weird things (like explicitly saying `undefined` or `failure`), but it sort of follows the grain of the code, and maybe it's useful in some cases.

athas:
I think the plotting code does not handle timeouts (or other missing data) correctly.

samestep:
> I think the plotting code does not handle timeouts (or other missing data) correctly.

Yeah I see that. Marking as draft for now, will fix and undraft as soon as I have the chance.

samestep (approved):
OK @athas I think this is ready now. I did have one question: I see that you added `--timeout 600` to `.github/workflows/build.yml` but not to `.github/workflows/nightly.yml`. Is that intentional?

athas:
No, that was not intentional.

samestep:
Alrighty, then yeah feel free to add that it to the nightly and then merge. Or I can do it once I get back to my laptop.

athas:
It is added. The ten minute timeout is pretty arbitrary, but we can always repaint that bikeshed in the future.

PR #292, opened 2025-02-27
Implement kmeans with finite differences.

athas:
This also rewrites finite.h to support higher order derivatives.

Marked as draft because it fails to validate on one of the (randomly generated) datasets. This may simply be a fundamental problem with applying finite differences to this problem.


athas:
This produces invalid results for some workloads, for reasons I think are fundamental to using finite differences for this problem. I tried fiddling a bit with the random workload generation to see if we could get lucky, but no dice. It may be possible if we take special care to generate well-clustered data where the bumping of the input doesn't change the assignment of points to clusters. For now, I have simply marked kmeans as `invalid` in `evals.txt`. I think it is still worth having this implementation, even if it doesn't validate in all cases.

PR #297, opened 2025-02-28
Implement GMM with JAX

athas:


python/gradbench/gradbench/jax/gmm.py
```diff
@@ -0,0 +1,55 @@
+import jax
+import numpy as np
+
+jax.config.update("jax_enable_x64", True)
```
samestep, line 4: I don't love mixing this line in among the imports; is there a better place to put it?

samestep: Is it possible to scope it using a Python `with` statement instead?

athas: The documentation emphasizes this must be done "at startup", whatever that means. There is a context manager version, but it is marked as experimental. Perhaps we can get away with doing this later down, though. I'm mostly concerned with whether it has to occur before the @jit annotations.

samestep:
Looks like JAX is once again the fastest.

samestep (approved):

PR #299, opened 2025-02-28
Use blue instead of red-green

### LINKED ISSUE #294: Website matrix colors could be confusing
As of #270, we use a red-to-green color gradient for the table of evals/tools on the website:

Image

The red color could give the incorrect impression that that tool is _incorrect_ for that eval, when really it just means that tool is slower than the fastest. We might want to consider instead using a different gradient, perhaps light blue to dark blue as suggested by @athas.

samestep:
Resolves #294.

Screenshot 2025-02-28 at 8 55 43 AM

### DIFF
--- a/packages/gradbench/src/App.css
+++ b/packages/gradbench/src/App.css
@@ -6,9 +6,16 @@ a:link {
   text-decoration: none;
 }
 
+.table {
+  background-color: #ccc;
+  padding: 5px;
+  width: fit-content;
+}
+
 .column-header {
   align-items: center;
   display: flex;
+  padding: 5px;
   writing-mode: vertical-lr;
 }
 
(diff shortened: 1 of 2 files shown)


athas (approved):
Eagerly awaiting the first tool slow enough to display some white for `hello`.

samestep:
@athas sorry that's not possible haha; the `hello` eval uses a simpler `Scorer` implementation that just always returns the same score:

https://github.com/gradbench/gradbench/blob/32baac06eb116fbe2164defce6892dbd07b26114/crates/gradbench/src/stats.rs#L44-L52

https://github.com/gradbench/gradbench/blob/32baac06eb116fbe2164defce6892dbd07b26114/crates/gradbench/src/stats.rs#L177-L184

PR #301, opened 2025-02-28
Limit Jax to a single thread

athas:
This is enforced via taskset in the Docker container itself, so it won't apply if you run Jax directly.

Surprisingly, Jax does not seem to have an obvious API to control parallel execution.


samestep (approved):
Elegant! We should also write documentation saying that currently tools are required to be single-core; I'll probably just do that when I get around to addressing #216.

PR #306, opened 2025-02-28
intermediary: show microseconds

### LINKED ISSUE #304: Intermediary does does not properly show runtimes lower than 1ms
```
$ ./gradbench run --eval './gradbench repo eval kmeans' --tool './gradbench repo tool jax'
  [0] start kmeans (jax)
  [1] def   kmeans                                424ms ✓
  [2] eval  kmeans::cost    k=10,n=1000,d=8     1.071 s ~         0ms evaluate × 19594 ✓
  [4] eval  kmeans::dir     k=10,n=1000,d=8     1.050 s ~         0ms evaluate × 11317 ✓
  [6] eval  kmeans::cost    k=10,n=1000,...     1.058 s ~         0ms evaluate × 12800 ✓
  [8] eval  kmeans::dir     k=10,n=1000,...     1.051 s ~         0ms evaluate × 8190 ✓
 [10] eval  kmeans::cost    k=10,n=10000...     1.100 s ~         0ms evaluate × 2545 ✓
 [12] eval  kmeans::dir     k=10,n=10000...     1.099 s ~         0ms evaluate × 1230 ✓
 [14] eval  kmeans::cost    k=100,n=1000...     1.034 s ~         0ms evaluate × 2818 ✓
 [16] eval  kmeans::dir     k=100,n=1000...     1.038 s ~         0ms evaluate × 1202 ✓
 [18] eval  kmeans::cost    k=10,n=10000...     1.145 s ~         0ms evaluate × 1756 ✓
 [20] eval  kmeans::dir     k=10,n=10000...     1.159 s ~         0ms evaluate × 1212 ✓
 [22] eval  kmeans::cost    k=100,n=1000...     1.037 s ~         0ms evaluate × 2611 ✓
 [24] eval  kmeans::dir     k=100,n=1000...     1.052 s ~         0ms evaluate × 1080 ✓
 [26] eval  kmeans::cost    k=100,n=1000...     1.173 s ~         2ms evaluate × 403 ✓
 [28] eval  kmeans::dir     k=100,n=1000...     1.254 s ~         5ms evaluate × 186 ✓
 [30] eval  kmeans::cost    k=1000,n=100...     1.093 s ~         1ms evaluate × 556 ✓
...
```

This is not terribly useful. One solution is to show microseconds instead, but then it can be annoying to compare microsecond with milisecond timing. We have the same problem with the ms/s changeover though, and that's in practice not a big problem.

athas:
This somewhat arbitrarily switches from writing milisecond-level numbers as 'xy.abc ms' to just ' xy ms' at 100 ms. This is supposed to ease the eyeballing of microsecond-level timings with milisecond-level timings, but there is going to be some kind of cut, unless we simply always print second-level timing with six digits (essentially using microseconds as our base unit).

This fixes #304.

### DIFF
--- a/crates/gradbench/src/intermediary.rs
+++ b/crates/gradbench/src/intermediary.rs
@@ -71,11 +71,11 @@ fn nanostring(nanoseconds: u128) -> String {
     let sec = ms / 1000;
     let min = sec / 60;
     if ms == 0 {
-        format!("{:2} {:2} {:3} μs", "", "", us)
+        format!("{:2} {:2} {:3}μs", "", "", us)
     } else if sec == 0 && ms < 100 {
-        format!("{:2} {:2}.{:03} ms", "", ms, us % 1000)
+        format!("{:2} {:2}.{:03}ms", "", ms, us % 1000)
     } else if sec == 0 {
-        format!("{:2} {:2} {:03} ms", "", "", ms)
+        format!("{:2} {:2} {:3}ms", "", "", ms)
     } else if min == 0 {
         format!("{:2} {:2}.{:03} s", "", sec, ms % 1000)
     } else if min < 60 {
@@ -448,17 +448,17 @@ mod tests {
 
     #[test]
     fn test_nanostring_0() {
-        nanostring_test("        0ms", Duration::ZERO);
+        nanostring_test("       0μs", Duration::ZERO);
     }
 
     #[test]
     fn test_nanostring_999_microseconds() {
-        nanostring_test("        0ms", Duration::from_micros(999));
+        nanostring_test("     999μs", Duration::from_micros(999));
     }
 
     #[test]
     fn test_nanostring_1_millisecond() {
-        nanostring_test("        1ms", Duration::from_millis(1));
+        nanostring_test("    1.000ms", Duration::from_millis(1));
     }
 
     #[test]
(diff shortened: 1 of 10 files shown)


samestep:
I understand that ideally it'd be nice to be able to see microsecond timings in the terminal but I don't like that this makes it hard to tell apart millisecond timings from second timings. Let me think more about this.

samestep (changes requested):
Yeah I don't find this to be an improvement over the status quo, because `1.234ms` looks too similar to `1.234 s`. I'm open to considering other alternatives, though.

athas:
> Yeah I don't find this to be an improvement over the status quo, because 1.234ms looks too similar to 1.234 s.

These numbers are however unlikely to occur adjacent in the output. In contrast, we see a whole lot of sub-milisecond timings for e.g. `gmm`.

samestep:
> These numbers are however unlikely to occur adjacent in the output. In contrast, we see a whole lot of sub-milisecond timings for e.g. `gmm`.

We do have sub-millisecond timings as you mention, but we also have plenty of timings in the range of one second to multiple minutes. As I said, I'm open to considering other alternatives that achieve a similar level of design consistency to the current printing scheme.

samestep:
Are we planning to come back to this with a different design? Or should I close this PR and/or close #304 as "not planned"?

athas:
I suppose you can close it, but it has become an increasing problem with the new evals, many of which have sub millisecond runtimes. My workaround is to instead gauge the number of runs done within a second.

samestep:
I'm happy to leave the issue open, but I'd be inclined to close this PR if we're not going with this specific design; unless you're still of the opinion that this is the right design and we should merge this after all?

PR #308, opened 2025-02-28
Put Python code for tools in `gradbench.tools`

samestep:
Been meaning to do this for a while. We already have the evals under `gradbench.evals`, so I think it makes sense to put the tools under `gradbench.tools`.


athas (approved):
This is a net improvement over the current state of things, but it always struck me as odd that tool-specific code does not live in `tools/`.

samestep:
> This is a net improvement over the current state of things, but it always struck me as odd that tool-specific code does not live in `tools/`.

Hmm it's been a while since I've thought about this. I like for our Python code to import other parts of our Python code using absolute imports like `from gradbench.whatever import stuff`, since I find Python relative imports to be extremely frustrating. But if there's a strong reason to move more code out of `python/gradbench` and into `tools` then we could consider changing it.

PR #309, opened 2025-03-01
Replace Black and isort with Ruff

samestep:
It's faster and replaces two tools with one. In a followup PR we can also use it to do [linting](https://docs.astral.sh/ruff/linter/).


athas (approved):
I don't have any opinion on Python formatters.

PR #310, opened 2025-03-01
Implement kmeans with adol-c

athas:
Strangely slow, despite using API functions that seem like they should do the efficient thing.


samestep (approved):
Thanks! Yeah I'd be curious to hear from an ADOL-C expert on why it's slow.

PR #311, opened 2025-03-02
Implement kmeans with cppad

athas:


athas:
I am pretty sure that this runs out of memory in CI (#278), which is incidentally fixed (in a somewhat crude way) by #290.

athas:
This might be due to 0553b2c4647badeba5dd3d788de7d3e050531711 where I made an incautious change to stop a crash.

samestep:
OK, I think it should be fixed in 66e0c2c1be373f7608504f3c412d390a0a804847.

athas:
You requested a review from yourself?

samestep:
Oh; lol I thought clicking that button would just clear the "changes requested" status so that merging would no longer be blocked, but I guess it didn't actually do that.

samestep (approved):

PR #314, opened 2025-03-02
tensorflow, pytorch, jax run.py: show error on import failure

athas:
Otherwise it is very annoying to debug errors that occur during the import.


samestep (commented):
Do you think it might make sense to print more information? See https://github.com/gradbench/gradbench/pull/288#issuecomment-2684995194, for instance.

athas:
Done.

samestep (approved):
Nice. Maybe we should put this `traceback` stuff in a utility function since now we call it in several places.

PR #315, opened 2025-03-02
Make TensorFlow implementations much faster.

athas:
They are now based on the "graph" versions from ADBench, which is often orders of magnitudes faster than the "eager" versions.

I will probably also port the others eventually.


athas:
I should also note that the TensorFlow BA implementation is still *awful*.

samestep (approved):
Looks good to me, and thanks for updating the links!

PR #337, opened 2025-03-04
Remove mygrad and autograd

athas:
In the medium term, there is no likelihood that these will actually be implemented, and it looks bad to have skeletal tools on gradben.ch.

In the long term, they are also not very interesting. They are too slow to handle large datasets, and their APIs are not too dissimilar to the other Python tools.

I am not opposed to mygrad or autograd in principle, so if someone aficionado wants to bring them back properly in the future, that is all fine with me.


samestep (approved):
Totally agreed. Thanks for pulling out the pruning shears!

PR #339, opened 2025-03-05
Undo wrong x-axis sorting on website

samestep:
Thanks @lecopivo for noticing this!

---

Before:

Screenshot 2025-03-05 at 4 03 36 PM

---

After:

Screenshot 2025-03-05 at 4 03 33 PM

### DIFF
--- a/packages/gradbench/src/Stats.tsx
+++ b/packages/gradbench/src/Stats.tsx
@@ -34,6 +34,7 @@ const makeSpec = ({
       x: {
         field: "workload",
         type: "nominal",
+        sort: null,
         axis: { labelAngle: -45 },
       },
       y: {


lecopivo:
I don't understand the ordering now. The workload is clearly indexed by a pair dimension and number of Gaussians,`(d,k)`. I think the data should be ordered lexicographically, ideally with an option to switch between `d` or `k` being dominant. Previously it was almost like that but the ordering on `k` was based on lexicographical ordering of strings rather than ordering of numbers.

For example previously it was clear that PyTorch clearly struggled with small `d`.

Maybe even better, you should be able show `k` plots showing scaling in `d` or `d` plots showing scaling in `k`.

samestep:
@lecopivo this PR just changes the website so it respects the ordering in which the eval sent the inputs to the tool; so, I don't think we should remove this `sort: null` from the website, and instead changes to this ordering should go in the eval code itself, in this case `python/gradbench/gradbench/evals/gmm/run.py`. This ordering has already undergone some iteration; for instance, see #283 by @athas. I'll leave it up to you two to decide what order the GMM eval should use for these workloads.

athas:
In the original ADBench graphs, the x axis contains the number of points in the problem (the product of n, k, and d). My ordering has the same spirit, but I maintain the actual parameters, because I found it very annoying to reconstruct a specific workload from the ADBench graphs (you had to find the one with the right product).

PR #340, opened 2025-03-05
Scilean gmm update

lecopivo:
Update of SciLean to newer version which simplifies some of the code in the GMM eval.


lecopivo:
Superseded by #354

Issue #225, opened 2025-02-04
CLI gives no "working" indicator while building before running


samestep:
As of #147 we have a `gradbench` shell script that uses [`cargo run --quiet`](https://doc.rust-lang.org/cargo/commands/cargo-run.html#display-options) to build the CLI before running it while not printing any extra output unless there are compile errors or warnings. Then, in #200, we added `repo eval` and `repo tool` subcommands that similarly use [`docker build --quiet`](https://docs.docker.com/reference/cli/docker/buildx/build/#options) when building evals or tools. These are great for decluttering the output when there is indeed nothing to build, but they're not great when there actually is stuff to build, because it results in no output at all during a potentially-long wait. It would be better to print build output _only when not everything is cached_.

PR #343, opened 2025-03-06
Print Cargo build progress in `./gradbench` script

samestep:
Addresses the Rust part of #225, but not the Docker part. I don't really want to merge this, though, because it's just so much nontrivial code. I wish that Cargo had the feature suggested in rust-lang/cargo#8743.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@
 Make sure to have these tools installed:
 
 - [Git][]
+- [Python][]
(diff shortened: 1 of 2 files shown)


samestep:
@athas does this look OK? The main thing I wanted to check with you on is now having `python3` as a dependency of the `./gradbench` wrapper script.

athas (approved):
Fine with me. `python3` is a harmless dependency.

PR #350, opened 2025-03-07
Change all Docker images to use /gradbench instead of /home/gradbench

athas:
This works with no downsides and maybe a purist can argue that it more appropriate since the Docker images have no 'gradbench' user. However, the real reason I am proposing this is simply because the Singularity container manager, which is relatively widely used in HPC environments, ends up removing /home/gradbench (under some important circumstances) because it has its own automation logic for handling /home.


samestep (approved):
Sweet! I thought about doing this originally but didn't have a specific reason one way or the other, so I chose arbitrarily. This new way is better anyways because it's shorter.

PR #353, opened 2025-03-07
Add CoDiPack tool

athas:
Only hello eval for now, but CoDiPack looks easy to work with, so I can probably do the others.


athas:
@MaxSagebaum, if you are willing to take a look at the quality of my code, that would be very appreciated.

athas:
I think CoDiPack is my favourite of the C++ tape tools so far. The performance is really impressive as well.

athas:
It is possible that my kmeans implementation does not use CoDiPack well, but it does work, although it is slow.

samestep (approved):
Amazing, thank you!

PR #354, opened 2025-03-08
SciLean implementation of k-means and lstm

lecopivo:
Implementation of k-means and lstm evals for SciLean


athas:
What's the challenge with LSTM?

lecopivo:
Ha no challenge, it works fine, I just can't do git. This branch was based on some ancient commit so  I was trying to rebase it on the recent main but half of lstm code disappeared.

lecopivo:
k-means was quite challenging. Mainly dealing with `let j := argMax fun j => -‖points[i] - centroids[j]‖₂²` which gives you index `j` depending on the variable you differentiate with `centroids`. 

My AD was designed to deal with functions between vector spaces but the type of indices do not form a vector space. I had to add a specialized rule allowing you to skip a let binding but you have to show that the value `argMax fun j => -‖points[i] - centroids[j]‖₂²` is constant on some neighborhood of  `centroids` at which you are taking the derivative.

lecopivo:
I don't understand the error in kmeans, all the runs finish successfully but then this error appears: 
```
expected exit code 2 (from outcome undefined) but got exit code 0
Error: Process completed with exit code 1.
```

lecopivo:
By updating SciLean I managed to break GMM :( I did a diff on the generated code and it is almost the same just few functions were renamed. Oh well, somehow I need to figure out what went wrong

athas:
You need to update the evals.txt file to mention kmeans and lstm.

athas:
It cannot be a real failure that `hello` crashes while much more complicated evals work fine.

lecopivo:
The issue is that hello does not seem to provide `min_runs` and `min_seconds` in the input JSON and right now I hard crash in such case. It should be an easy fix

samestep:
Yeah @lecopivo your implementation shouldn't assume that the `"input"` field is consistent across evals or even across functions within the same eval; ideally you'd handle the input for different evals independently, and handle shared cases like `"min_runs"`/`"min_seconds"` by defining helper functions and calling them within those specific evals.

lecopivo:
I think I take exactly this approach, just that the utility function `wrap` works of gmm, lstm, kmeans but not for hello.

athas:
I suggest looking at how the Haskell and OCaml implementations do this. It comes out quite naturally.

lecopivo:
Everything works now!

PR #357, opened 2025-03-11
codipack: reset the tape before each evaluation

athas:
Based on a discussion on the Discord, I decided to make the CoDiPack implementations of GMM and LSTM explicitly reset the tape (and indicate the inputs) during each evaluation. This is due to uncertainty about how much work CoDiPack can otherwise reuse across evaluations - the first executions are *much* slower than the rest.

CoDiPack is still pretty fast overall, but this does make it slower on some workloads.


samestep (approved):
Sounds good to me. Let me know what Max says when you ask him about this at EuroAD!

PR #359, opened 2025-03-12
ba: enable all workloads

athas:
This one is a bit annoying to actually run because the BA Jacobians are large and take a long time to transfer via JSON. Since the BA input is actually replicated by the tool (not documented!), maybe it is worth not actually sending the entire Jacobian, since it is very redundant. It is not obvious to me how to isolate the replicated part, however.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -228,6 +228,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Increase disk space
+        uses: ./.github/actions/space
       - name: Install CLI from artifact
         uses: ./.github/actions/cli
       - name: Download all log artifacts
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -228,8 +228,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Increase disk space
-        uses: ./.github/actions/space
       - name: Install CLI from artifact
         uses: ./.github/actions/cli
       - name: Download all log artifacts
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -178,6 +178,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+      - name: Increase disk space
+        uses: ./.github/actions/space
       - name: Install CLI from artifact
         uses: ./.github/actions/cli
       - name: Download all log artifacts
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -178,8 +178,6 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Increase disk space
-        uses: ./.github/actions/space
       - name: Install CLI from artifact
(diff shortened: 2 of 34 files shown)


samestep:
Regarding the JSON transfer: as I mentioned in https://github.com/gradbench/gradbench/issues/215#issuecomment-2632104874, it's quite possible that we can speed up our JSON processing in general, at least in the Rust part. I just haven't spent time looking into this yet because it hasn't seemed like a bottleneck so far.

athas:
It may well be that the slow part is the validation. I've been meaning to take another look at that comparison function and perhaps speed it up.

samestep:
Another possibility to speed things up could be to rewrite the evals (and thus the validation) in Rust. 😛

athas:
Fine with me. And maybe switch the input/output data formats to something more efficient than JSON.

samestep:
Wait huh? I assume you're being sarcastic?

athas:
It's not something I consider urgent (either of them), but in an ideal world I would do both of these. JSON and Python are far from the worst choices, but they are not literally the best.

samestep:
Sure, but switching from Python to Rust doesn't affect the tool implementations at all. Switching from JSON to something else then means that every single tool needs to have an implementation for that format. To me it feels like it'd impose too much of a barrier, but perhaps I'm missing something?

athas:
Ugh, these logs are now so large that they don't even fit in the temporary filesystem.

samestep:
Let's see if it works after we add more disk space.

athas:
Another option is to prune the input/output fields after downloading. They are not needed for the stats.

samestep:
Sure; it's just not the first thing I'm trying because downloading them all with one `gh` command is very convenient.

athas:
Looks like that wasn't enough.

samestep:
Yeah. Could try deleting even more stuff to add even more disk space. Or could try something different.

athas:
I would expect the `ba` log files to compress very well because of the duplication. In general, maybe it is worth compressing all the log files and automatically decompressing in `gradbench`.

samestep:
All the log files are already automatically compressed when they're uploaded as artifacts in GitHub actions. For instance, if you look at the workflow run for commit eec90edc4b96f2a9725e700b5099baab4caefe48, you'll see that the size of each BA log artifact is roughly 290 megabytes; that's the _compressed_ size: https://github.com/gradbench/gradbench/actions/runs/13812867690

athas:
Yeah, the uncompressed size is 4.8GiB. This won't fly. If we want un-pruned log files, then I'll have to modify the `ba` eval to not send the full Jacobian. I'll give it a shot tomorrow.

athas:
I have a solution for this in #368 which I will merge when it works, but I do not think it is realistic that we can always fit all log files (including input/output) in a runner VM at the same time. The `ba` outouts have an average size of 354MiB, which is not at all unreasonable for a numerical benchmark. This will get worse as we add more evals and tools in the future.

samestep:
Yeah, fair enough; seems like we need to rethink something. Could you open an issue so we can discuss the different possible solutions here?

athas:
Nicely smooth graphs here.

samestep (approved):

PR #368, opened 2025-03-13
ba: deduplicate Jacobian

athas:
This modifies the ba eval to not require transmitting the repeated parts of the Jacobian.

The purpose of this is to significantly shrink the size of the ba log files.


athas:
Note that although I have updated the documentation, I still consider the `ba` documentation very mysterious. In particular, it does not mention the duplication of the input, and neither does any part of the original ADBench documentation. It is not understandable without reading the code.

samestep:
@athas would it be feasible to update the BA documentation in this PR while you're looking at it, so we don't have to come back later to make it actually decent?

athas:
No, because I don't fully understand the motivation for what is going on, or exactly why it makes sense.

PR #371, opened 2025-03-13
ht: stick to one workload class

### LINKED ISSUE #367: On ht
The primal graph for `ht` currently looks like this:

Image

While the underlying data is correct, this is clearly not what a graph should look like. The reason for this graph is that the `ht` workloads are actually four qualitatively different ones.

1. First, there are workloads corresponding to two hand models, called "small" and "large" - according to the ADBench paper, these are supposed to imitate "online" and "offline" applications of hand tracking.
2. Second, each model has "simple" and "complicated" workloads, which in many of the tools are implemented by separate code. The "simple" one leaves out the part of the objective function (and input) that contributes to the sparse part of the Jacobian (specifically, the `us` vector of the input is empty). The difference in runtime is insignificant. The ADBench paper does not motivate why it is worth it to have both - the "simple" one seems like it is just *wrong*, as it doesn't actually solve the full problem. None of our tools have any particular difficulty expressing the "complicated" version, but perhaps that is the case for some of the tools that we did not port (maybe some of the MATLAB ones).

I can think of the following options, not all of which are mutually exclusive:

1. Modify the graph code so that it does not draw a line between the datapoints from different classes of workloads.
2. Remove the "simple" workloads as I don't think they contribute anything.
3. Split `ht` into two distinct evals: `ht-simple` and `ht-complicated`. This is essentially what ADBench does, although in a somewhat half-hearted manner.
4. Remove all but the "complicated_large" workloads.

Option (4) is clearly the easiest to implement. It's also worth nothing that even the "large" workloads have pretty small primal runtimes (single-digit miliseconds) on most tools, and the Jacobians are less than two orders of magnitude slower.

Option (1) maximises information, but requires bespoke modification to the graphing code, and I'm quite uncertain that this added "information" is actually useful.

athas:
I'm indecisive so I retained the ability to run the different classes with explicit arguments passed to the eval.

Fixes #367.


samestep (approved):
Sounds good to me!

samestep:
The charts look a lot more reasonable now.

PR #375, opened 2025-03-13
Freshen up Nix setup

athas:
* Moves shell.nix to the repository root, which is more idiomatic.

* Pins Nixpkgs with niv, in order to ensure reproducibility.

* Adds a few more documenting comments.


samestep (approved):
Nice. I look forward to using this once I've learned enough and gotten my own machine's Nix setup far enough!

samestep:
@athas why did you choose to use `niv` instead of flakes?

athas:
I'm more familiar with Niv. It also works by default, while Flakes is still an experimental option that has to be turned on. I wanted a setup that would work immediately even for people who don't know more about Nix than how to install it.

PR #377, opened 2025-03-13
lstm: generate data on the fly instead of using data files

athas:
This is more flexible, and the data generation is very fast anyway.

To demonstrate this great new flexibility, I added some more default workloads. This should also make the graph look more interesting.


samestep (approved):
Sweet, I love it!

Issue #363, opened 2025-03-12
No easy way to generate charts locally


samestep:
As of #270 we use [Vega-Lite](https://vega.github.io/vega-lite/) to generate charts dynamically on the website. However, there's currently no straightforward way to generate those charts locally, which makes it difficult to run experiments involving different tools or custom eval configurations.

In #342 we modified the `gradbench stats` command to also emit an SVG file, which goes part of the way toward resolving this issue, but that code is essentially duplicated between the CLI and the website, making maintenance a bit more error-prone as in #290. And of course, this is only the most basic chart; the more detailed plots are still only implemented in JavaScript.

One possible solution to this could be to move all the plotting code entirely into Rust, and then deduplicate the implementations on the website by compiling that portion of the Rust codebase to WebAssembly to reuse in the website. On the one hand, this would take some more work since we would no longer be able to lean on Vega-Lite, but on the other hand, the set of charts we need is pretty constrained, and there might not be a way to resolve some of our existing issues like #296 in Vega-Lite anyways, so this flexibility might actually beneficial.

athas:
Could we not run Vega locally, in Rust, to produce an SVG? While I have never actually used Vega for anything, the idea of a visualisation grammar is in principle appealing.

samestep:
@athas sure, that's another option: assume that Bun is installed and `bun install` has been run to populate `node_modules`, and then it should be possible to run Vega as a script.

PR #378, opened 2025-03-14
New evals: particle, saddle

athas:
These are benchmarks from the 2016 paper *Efficient Implementation of a Higher-Order Language with Built-In AD*:
https://www.bcl.hamilton.ie/~qobi/ad2016-benchmarks/

For some reason I am unable to find the paper that describes what precisely they do, but from what I remember, 'particle' is about simulating the movement of a particle in an electric field, and 'saddle' is about finding a saddle point for some function. Both make use of the Euler method for doing so.

The compelling thing about these problems is that they involve nested AD, and not just in the simple way needed by kmeans for computing hessians. No, these things optimise objective functions that themselves contain instances of AD (in the case of 'saddle', this is a nested solver). The 'particle' eval even has unbounded 'while' loops in the differentiated part. These ones are really looking for a fight. Both have two levels of AD.

This PR is marked as a draft for various reasons, but the main one is that these benchmarks don't fit real well in GradBench's current model, so we need to figure out how to map them to the module/function/workload hierarchy. For example, there is no meaningful notion of a "primal" function here for which we can measure AD overhead, like we do for all of the extant evals. But it gets worse.

Each of the evals have two levels of AD, and each level can be implemented with forward and reverse mode. Siskind and Pearlmutter report performance numbers for each of the four combinations here: https://engineering.purdue.edu/~qobi/stalingrad-examples2009/

I have chosen to model these four variants as four "functions", within the same module.

In this PR, I only have a single tool implementation (Futhark), and I implement each of the four variants for each of the two evals. This works fine in Futhark, but I suspect some tools will have trouble expressing some of these cases - doubly nested reverse mode with 'while' loops is *not* a trivial exercise. How will we handle the case where a tool can only express a subset of these four combinations?

The functions have only a single workload each, taken from Siskind and Pearlmutter's code distribution. In general, it seems scalability across workloads was not a priority for Siskind and Pearlmutter, and I arguably agree - there's not really any meaningful "input size" you can dial here. One unfortunate consequence is that these problems are vey sequential, but I think that's only a problem due to my own research interests.

As alluded to above, these are complex uses of AD, and we should not expect to easily obtain many tool implementations for these evals. I am pretty sure Enzyme can do this (and I will give it a shot), and maybe some of the other C++-tape-based ones (though I already dread the template error messages). I was a bit surprised that Futhark managed to handle it well. I don't think we can use any of the Siskind and Pearlmutter code, as most of it involves implementing AD facilities from scratch in the various languages, which is not appropriate for GradBench. (Also, most of these languages are just different Scheme implementations).


athas:
Oh yeah, and any plotting code for these will need to be completely bespoke. Probably something like a grouped bar chart, with a group per function, and a bar per tool.

athas:
The benchmarks are properly described here: https://engineering.purdue.edu/~qobi/papers/ad2008.pdf

evals/particle/README.md
samestep, line 1: This contextual and mathematical documentation is great, thanks! An eval's README should also contain interface info for tool implementers, though:

- Explicitly document that this eval defines a module named `particle`.
- Explicitly document that the `particle` module contains functions named `rr`, `ff`, `fr`, and `rf`, all with the same input/output interface.
- Explicitly document that the `"input"` field `extends Runs` and has a field named `"w"` of type `number`.
- Explicitly document that the `"output"` is a `number`.

evals/saddle/README.md
samestep, line 1: Basically the same comment as for `evals/particle/README.md`:

- Explicitly document that this eval defines a module named `saddle`.
- Explicitly document that the `saddle` module contains functions named `rr`, `ff`, `fr`, and `rf`, all with the same input/output interface.
- Explicitly document that the `"input"` field `extends Runs` and has a field named `"start"` of type `[number, number]`.
- Explicitly document that the `"output"` has type `[number, number, number, number]`.

samestep (commented):
Really nice work on this, thanks for bringing these into GradBench. Broadly I'm in favor of this, so we just need to figure out the details.

> This PR is marked as a draft for various reasons, but the main one is that these benchmarks don't fit real well in GradBench's current model, so we need to figure out how to map them to the module/function/workload hierarchy. For example, there is no meaningful notion of a "primal" function here for which we can measure AD overhead, like we do for all of the extant evals.

This is actually all fine, since that idea of a "primal" function isn't a core GradBench concept: it's just part of the way we score and plot data for a few of our existing evals. We call this [`ScorerClassic`](https://github.com/gradbench/gradbench/blob/e3a76c38ca7d5acad6c61c725a8508910001af0f/crates/gradbench/src/stats.rs#L121) in our Rust code, and it's [explicitly demarcated](https://github.com/gradbench/gradbench/blob/e3a76c38ca7d5acad6c61c725a8508910001af0f/crates/gradbench/src/stats.rs#L201-L208) as only one possible scorer, with other options available. So there's no issue here, as you can see by the fact that these new evals pass through CI just fine: there's no special scorer defined for them, so they get [the default scorer](https://github.com/gradbench/gradbench/blob/e3a76c38ca7d5acad6c61c725a8508910001af0f/crates/gradbench/src/stats.rs#L45) instead. We can easily add a new kind of `Scorer` which applies to `particle` and `saddle`.

> In this PR, I only have a single tool implementation (Futhark), and I implement each of the four variants for each of the two evals. This works fine in Futhark, but I suspect some tools will have trouble expressing some of these cases - doubly nested reverse mode with 'while' loops is not a trivial exercise. How will we handle the case where a tool can only express a subset of these four combinations?

This is the real issue. I see a few possible solutions:

1. Keep it as you currently have it, with four functions corresponding to the four variants. Tools that can't do all of them will just not be able to successfully complete the whole eval; we already have plenty of tools that fail on other evals for other reasons.
2. Split each into four different evals, with one function each, so a tool can implement only the subset that works for it. This may be "morally" the correct thing to do, but I don't really like it. It feels unnecessarily "verbose" in a way.
3. Just have one function in the `saddle` eval and one function in the `particle` eval, which doesn't specify _how_ the tool which of the four strategies the tool should use to implement that function, since that's just an implementation detail and has nothing to do with the mathematical definition of the function. Select among those implementations in other ways (e.g. CLI flags on the tool itself).

Of these, (2) is my least favorite, (1) seems fine, and (3) is my preference. In my opinion, the GradBench evals should define _mathematical functions_, not dictate specific strategies and implementation details for tools to implement those functions. And if that makes these evals less useful because then a tool could successfully implement the eval by just doing one of the four strategies, that's the fault of the eval, not the tool. If we want to demonstrate in GradBench which tools can and can't nest autodiff in various ways, we should have evals that _require_ you to nest autodiff in that specific way, by making it so that nesting in one of the other three ways would be prohibitively slow.

> Oh yeah, and any plotting code for these will need to be completely bespoke. Probably something like a grouped bar chart, with a group per function, and a bar per tool.

Agreed. As I mentioned above, we already keep `Scorer` implementations separate per eval; the fact that we don't already do this in the website too is just because I didn't get around to implementing it yet. Going forward we really should have the plotting code dispatch based on the eval; this also relates to #363.

athas:
I'll write proper protocol documentation once we decide on the structure of the evals.

athas:
Of your three options, I am strongly in favour of (1). While I normally support tools doing whatever works best for a given eval, an important point of `particle` and `saddle` is to investigate the difference in F/R behaviour. Even comparing just Futhark and Enzyme, it is very interesting that it is not consistent which combination is best. It would be a shame to lose this richness. For colouring the square on gradben.ch, you can always just pick the best combination as the "representative" number for a given tool.

athas:
The way these inputs would be handled by any tool would be by a case that dispatches to the specified function based on the input. It is awkward to handle cases where you need to prepare special data structures, e.g a tape, based on what the input looks like. This is unlike how any of our tools handle any existing eval. I think there is no value in sneaking in a function through an input. It also does not simplify the problem of handling missing variants. It does make it slightly harder to extract information from the logs, as it is a bit easier to filter based on the function, which is always present, but it is not a big deal. The output from the intermediate looks a bit nicer with variations in function name rather than workload name, although that is subjective.

samestep:
> The way these inputs would be handled by any tool would be by a case that dispatches to the specified function based on the input. It is awkward to handle cases where you need to prepare special data structures, e.g a tape, based on what the input looks like. This is unlike how any of our tools handle any existing eval.

I really don't understand this point. Just to see what the code would look like, I went and implemented it myself, and it seems totally fine: 85a72578d05a3bb43396c51b7788aff15a34d6f9

---

> I think there is no value in sneaking in a function through an input.

I already described the value in my earlier message:

> The reason I want this over the four-functions approach is that these `inner`/`outer` parameters are _mathematically_ meaningless (it's the same function being computed regardless), but they are _computationally_ meaningful, similar to `min_runs` and `min_seconds`. And you could imagine having other computationally meaningful parameters as well, in which case it doesn't seem right to privilege these ones by putting them in the function name while others would just go in the parameters anyway. (Also, this is just more obvious than trying to remember the difference between `fr` and `rf`.)

Case in point: it took me about as long to just figure out which one was `fr` and which one was `rf` as it took me to just write the code once I knew which was which.

---

> It also does not simplify the problem of handling missing variants.

Sure, my answer for that is still the same I wrote above:

> Tools that can't do all of them will just not be able to successfully complete the whole eval; we already have plenty of tools that fail on other evals for other reasons.

---

> It does make it slightly harder to extract information from the logs, as it is a bit easier to filter based on the function, which is always present, but it is not a big deal. The output from the intermediate looks a bit nicer with variations in function name rather than workload name, although that is subjective.

Seems like we both agree that these aren't significant issues.

athas:
> I really don't understand this point. Just to see what the code would look like, I went and implemented it myself, and it seems totally fine: [85a7257](https://github.com/gradbench/gradbench/commit/85a72578d05a3bb43396c51b7788aff15a34d6f9)

You did have to rewrite some of the supporting code and add new concepts, which is my point: this wasn't necessary with the multiple-functions approach. There is also now a discrepancy between the function at the protocol level and the function name that gets passed to the C++ `main()`. I see no value in losing this consistency. It can work, but it's extra moving parts for which I am not convinced we gain anything.

Regarding the FF/FR/RF/RR terminlogy, that is taken directly from the original benchmarks. We can change it to something else if we want - that is an orthogonal issue. But note that these longer names have a significant risk of not being fully shown with the current `gradbench` printing format. There is value to terse names, as there's a whole bunch of information we want to fit in a single line.

samestep:
I think you may be over-prioritizing short-term ease of implementation and leaking implementation details, whereas I am attempting to choose the best interface design irrespective of those.

> You did have to rewrite some of the supporting code and add new concepts, which is my point: this wasn't necessary with the multiple-functions approach. There is also now a discrepancy between the function at the protocol level and the function name that gets passed to the C++ `main()`.

I think it's too easy to use this sort of argument to justify overfitting to only the set of things we've considered in the past. Part of the value in adding more benchmarks is that as we do so, we'll realize that some of the patterns we took advantage of before were not universal, and so we should adjust to choose better abstractions going forward. Just because my prototype Enzyme implementation for this proposed interface has this discrepancy doesn't mean that's necessary; it might just mean we want to do a bit of refactoring after this PR is merged.

> Regarding the FF/FR/RF/RR terminlogy, that is taken directly from the original benchmarks. We can change it to something else if we want - that is an orthogonal issue. But note that these longer names have a significant risk of not being fully shown with the current `gradbench` printing format. There is value to terse names, as there's a whole bunch of information we want to fit in a single line.

I agree that this is an orthogonal/minor point, but it's worth noting that my approach avoids this entirely: we use the `description` to print a short identifier for the console log, and use more descriptive field names to feed to the tool.

athas:
More points: the saddle.py/particle.py logic has to be duplicated and maintained for every eval.

And regarding the motivation: for this benchmark, these are *not* completely black box mathematical functions that can be implemented any which way. They *must* be implemented using nested solving, or else this benchmark is pointless. You can certainly solve the problems with other algorithms, e.g. a four parameter gradient descent for `saddle`, but that is not the specification. If we cannot accept that these are a kind of toy problem with some arbitrary implementation constraints, then we should simply reject them from GradBench.

samestep:
> More points: the saddle.py/particle.py logic has to be duplicated and maintained for every eval.

Not sure exactly what you're referring to here; is this relevant? adad4bb6109beaebfc06d2e04e83e561b0636080

> And regarding the motivation: for this benchmark, these are _not_ completely black box mathematical functions that can be implemented any which way. They _must_ be implemented using nested solving, or else this benchmark is pointless. You can certainly solve the problems with other algorithms, e.g. a four parameter gradient descent for `saddle`, but that is not the specification. If we cannot accept that these are a kind of toy problem with some arbitrary implementation constraints, then we should simply reject them from GradBench.

I agree. The choice of simply rejecting these evals sounds more like the option (3) that I listed above, which it sounds like you don't prefer, and as I've written in this thread, I'm inclined to agree with you that something more like choice (1) is better. I'm simply proposing an alternate interface for what is essentially the same as the original choice (1).

athas:
I don't have any issue with changing the implementation to fit new needs. I rewrote it plenty of times already, after all. However, a significant bar must be cleared in order to add complexity (even just treating lines of code as a proxy for complexity). I think the following is the best argument for putting the ff/rr stuff in the input field:

* All variants are expected to compute the same result (perhaps modulo some numerical instability, although that does not seem to actually occur in practice).

But it is important that these are not just specified as *mathematical* functions - according to the specification they must be implemented in a certain way. This to me is different from what we put in the input field, which I see as mathematical values that can be processed in whichever way is appropriate.

1. The `min_runs`/`min_seconds` fields don't quite fit this model, but a tool can largely ignore these fields as far as correctness goes, and they mostly exist because there is no other convenient place to put them).
2. We still assume some kind of "good faith" on behalf of implementations, e.g. don't just have a table of the results for the known workloads, and *actually do perform AD* rather than just just having a manually differentiated function. But this is still less restrictive than the imposition of a specific algorithm that is done by `particle`/`saddle`.

When it comes to plotting or any kind of analysis, there is no great difference between the input and function name approaches. Nor when it comes to graphing. If we ever add more workloads to each variant, then it becomes somewhat annoying that the input describes both a variant and a normal workload point, because they can't all be directly compared in the same way. Then it is useful that the "function" is one axis and the input is another. (This is a bit of a strawman argument, as I don't think we will want more workloads for these evals, as you will see when I improve the documentation.) When it comes to the current C++ infrastructure, then it is *very* useful that operationally different algorithms (e.g. ff vs rr) are different function names, because that results in different classes being instantiated, which can then have bespoke data structures depending on the approach. Some of our tools will require very different code for these different versions. This can be done by synthesising function names in the Python code, as in your branch, but it adds another level of indirection that I do not believe is justified by what it gives us.

I will write up protocol documentation based on my current design.

samestep:
I think there's an important point I've been making that I haven't seen you respond to yet, so I'd like to highlight that. Specifically, in your most recent message, you wrote:

> I think the following is the best argument for putting the ff/rr stuff in the input field:
>
> -  All variants are expected to compute the same result (perhaps modulo some numerical instability, although that does not seem to actually occur in practice).

This is only part of what I've been saying. Here's the rest of it, which I don't think you've responded to yet:

> The reason I want this over the four-functions approach is that these `inner`/`outer` parameters are _mathematically_ meaningless (it's the same function being computed regardless), but they are _computationally_ meaningful, similar to `min_runs` and `min_seconds`. And you could imagine having other computationally meaningful parameters as well, in which case it doesn't seem right to privilege these ones by putting them in the function name while others would just go in the parameters anyway. (Also, this is just more obvious than trying to remember the difference between `fr` and `rf`.)

In other words, `inner` and `outer` are _already_ not the only parts of the `"input"` that dictate computation rather than just the function. And you pointed this out in your most recent message:

> 1. The `min_runs`/`min_seconds` fields don't quite fit this model, but a tool can largely ignore these fields as far as correctness goes, and they mostly exist because there is no other convenient place to put them).
> 2. We still assume some kind of "good faith" on behalf of implementations, e.g. don't just have a table of the results for the known workloads, and *actually do perform AD* rather than just just having a manually differentiated function. But this is still less restrictive than the imposition of a specific algorithm that is done by `particle`/`saddle`.

These align with what I've been trying to say. I'm contending that it's quite reasonable to expect that in the future we'll want to add more things to the `"input"` that affect the computation rather than just the mathematical function (since we've already done that with `min_runs`/`min_seconds` and now we're doing it again), and if/when that happens, it will be more consistent to have all these computation-specifying parameters live alongside each other in the `"input"`, rather than having some be privileged as part of the function name and others not.

As you mentioned here, the tool can ignore `min_runs`/`min_seconds` and still give the correct answer. The same is true for `input`/`output`. It's also the case that if the tool does ignore `min_runs`/`min_seconds`, we won't get good benchmark data. This is also true for `input`/`output`. So they really do seem like the same kind of thing, which is why I'm arguing that we should represent them in the same way.

To put it more concisely: do you really expect that in the future we're not going to add any more computation-specifying parameters, despite the fact that we've done so in the past?

athas:
You are right that I didn't address that directly, and I apologise for that. I will do so now, although it does not change my opinion for this eval specifically.

> The reason I want this over the four-functions approach is that these inner/outer parameters are mathematically meaningless (it's the same function being computed regardless), but they are computationally meaningful, similar to min_runs and min_seconds. And you could imagine having other computationally meaningful parameters as well, in which case it doesn't seem right to privilege these ones by putting them in the function name while others would just go in the parameters anyway. (Also, this is just more obvious than trying to remember the difference between fr and rf.)

These `saddle`/`particle` evals are specifically not about mathematical functions, but particular algorithms, with a (relatively) well specified operational behaviour. This is not the only way to define benchmarks (or even the best one in general), but it is the one that Siskind and Pearlmutter decided for these - and I think it does make sense to have this; at least the results are not what I expected. It is true that it is not scalable to put the computational variants into the function name in general, but these particular evals will likely *never* change - the point of them is that they are an adaptation of benchmarks that already exist out there. I would probably not specify them Siskind and Pearlmutter did (in particular I would not have them be quite as sequential), but they are fine for what they are.

I am uncertain if we will have benchmarks that have even more computational variants than these. I am not aware of any, and I am reluctant to specify any myself, as they dramatically increase the implementation effort.

> To put it more concisely: do you really expect that in the future we're not going to add any more computation-specifying parameters, despite the fact that we've done so in the past?

For `saddle`/`particle`, I am certain we will not add any more variants. For future evals, maybe. But in that case I think the needs of those future evals should inform our design. If the study of semantically equivalent "variants" becomes particularly interesting to GradBench for some reason, it could even be argued that it should be reified as an actual protocol-level concept, but to be honest I think we actually have a good mechanism for that: you can have different modules that present the same interface, but with with different implementations. This is also more general, but slightly heavy on the boilerplate currently, so I'd rather not go that way until we have a particularly compelling use case, and `saddle`/`particle` actually work really well just by exposing four functions.

samestep:
Makes sense, thanks. I still prefer my version, but since it's now clear that you've considered all the points I consider important, I'll yield to your design.

athas:
Thanks. To celebrate I have documented `particle`. I am a little bit worried about my own lack of skill at succinctly describing the required mathematical approach. Honestly, I only understood myself what was going on after reading Siskind and Pearlmutters own code (their Haskell implementations in particular, but they're all pretty nice).

athas:
I want to do two more implementations before marking this as ready:

* Finite differences, which I think will actually work reasonably well and perhaps undermine the AD story, but so be it.
* An implementation in at least one tape recorder.

samestep (commented):
I'll finish reviewing this tomorrow.

samestep (approved):
Amazing, LGTM! Thanks again for adding these.

PR #379, opened 2025-03-14
Update SciLean version

lecopivo:
I did a bit of profiling on `kmeans` and improved the performance.


athas:
How big is the speedup?

lecopivo:
Quite nice speed up, around 10x. 

For `kmeans` objective, I can write Lean code that is as fast as C code but idiomatic SciLean code is still 2x slower, but the code looks so pretty:
```
∑ᴵ i, minᴵ j, ∑ᴵ l, (points[i,l] - centroids[j,l])^2
```
I still should do more profiling on the derivative. It is reasonably fast already but I don't see a reason to get similar result as `manual`.

lecopivo:
It was quite an interesting experience, I have learned to use `perf`. The key things I had to do:
- switch from internal to external iterators
- stop using arbitrary sized numbers `Nat` and switch for 64-bit numbers when indexing arrays. Here I likely break consistency, but that would come up only if you work with arrays that do not fit into the memory ... so who cares.
- inline bunch of functions. Mainly some higher order functions that should be converted to raw loops.
- I have bunch of types that are `ByteArray`+proofs. Under certain conditions you can convert one to the other just by passing the byte array and reprove few things. This should be identity at runtime, but it was causing performance issues. I was forced to just do unsafe cast from one to the other. This had to be a big in Lean compiler.

lecopivo:
Ok SciLean is working but CI fails because of TensorFlow, not sure what to do about that.

PR #383, opened 2025-03-17
Flesh out Nix config enough for VS Code users

samestep:
Most recently, #375 improved the Nix config a fair bit. This PR adds a couple missing things necessary for developers using VS Code, with a particular focus on workflows involving the [direnv](https://marketplace.visualstudio.com/items?itemName=mkhl.direnv) VS Code extension, and documents the Nix config in `CONTRIBUTING.md`.

- [We recommend](https://github.com/gradbench/gradbench/blob/5c168fd049339271c0ee270a8ba6fe1ffe70c0a9/.vscode/extensions.json#L6) the [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) extension, which gives `Error: Cannot find module 'prettier' from '/home/sam/github/gradbench/gradbench'` if `bun install` has not been run, so this PR adds `pkgs.bun`.
- [We recommend](https://github.com/gradbench/gradbench/blob/5c168fd049339271c0ee270a8ba6fe1ffe70c0a9/.vscode/extensions.json#L12) the [rust-analyzer](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer) extension, which gives `` ERROR can't load standard library, try installing `rust-src` `` if `RUST_SRC_PATH` is not set, so this PR sets it.
- [We configure](https://github.com/gradbench/gradbench/blob/5c168fd049339271c0ee270a8ba6fe1ffe70c0a9/.vscode/settings.json#L13) `"rust-analyzer.check.command": "clippy"`, which gives `` error: no such command: `clippy` `` if Clippy is not available, so this PR adds `pkgs.clippy`.

Finally, this PR creates an `.envrc` file telling [direnv](https://direnv.net/) to activate `shell.nix`, and also adds the resulting `.direnv` folder to `.gitignore`.

While I was at it, I also removed Git as a dependency from the CLI (since it wasn't really necessary), and made our `./gradbench` script a bit more robust to the cwd.


CONTRIBUTING.md
```diff
@@ -38,6 +38,8 @@ Other tools that are optional but useful:
 - [uv][]
 - [Make][]
 
+If you're using [Nix][] and [direnv][], everything you need is in the `shell.nix` file at the root of this repo, which will be automatically activated once you've entered the directory if you've run `direnv allow` at least once.
```
athas, line 41: Strictly speaking not *everything*. There may be tools that have dependencies not provided by `shell.nix`. In particular, I couldn't figure out what SciLean needs.

athas (approved):

CONTRIBUTING.md
samestep: Good point, I'll clarify that comment.

PR #392, opened 2025-03-20
Explicit tool colors when plotting

### LINKED ISSUE #346: Website chart legends duplicate colors
This relates to #295. As an example:

Image

This shows that Vega-Lite's default color scheme starts cycling after 10 colors.

athas:
These colors are pretty arbitrarily set, but easy to change. It still doesn't make the graphs exactly easy to read, but at least they are unambiguous. When I could not find a theme color on a tool home page, I tried writing the tools name in 1337 speak.

Fixes #346.

### DIFF
--- a/packages/gradbench/src/Stats.tsx
+++ b/packages/gradbench/src/Stats.tsx
@@ -8,7 +8,7 @@ import "./Stats.css";
 // some are unfortunately a little close to each other.
 const colors = [
   ["adept", "#d8702f"],
-  ["adolc", "#0d1117"],
+  ["adol-c", "#0d1117"],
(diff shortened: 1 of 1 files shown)


athas:
In an ideal world, I think it would be nice for each tool to have a file `metadata.json` that contains its preferred color, display name, and other such things. But I couldn't figure out how to make that information to the website build system, and this PR is a decent intermediate step.

athas:
ADOL-C now has an ugly colour, but it is distinct. The ADOL-C maintainers can always suggest a prettier colour.

samestep (approved):

PR #393, opened 2025-03-20
Put the derivative chart above the ratio chart

samestep:
I think it gives a better overall view of which tools are fast and which are slow.


athas:
I think this is mostly a signal of which tools happen to be implemented in C++, but I am not very concerned with the order.

athas (approved):

PR #394, opened 2025-03-21
Use uv in `shell.nix`

samestep:
I don't see a reason to list all the Python packages separately via Nix, resulting in different versions from `uv.lock`. @athas I seem to remember you saying there was some reason not to do this, though; can you remind me what that is?

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -23,10 +23,11 @@ in pkgs.stdenv.mkDerivation {
   name = "gradbench";
   buildInputs = [
     pkgs.bun
-    pkgs.niv
     pkgs.gh
-    pkgs.uv
+    pkgs.niv
     pkgs.nixfmt-classic
+    pkgs.python311
+    pkgs.uv
 
     pkgs.futhark
     pkgs.enzyme
--- a/shell.nix
+++ b/shell.nix
@@ -1,8 +1,5 @@
 # shell.nix for use with Nix, in order to run GradBench locally.
 #
-# May get out of sync with the uv setup, but hopefully it is not too
-# difficult to maintain manually.
-#
 # The Nixpkgs snapshot is pinned with niv. Use
 #
 #  $ niv update nixpkgs -b nixos-unstable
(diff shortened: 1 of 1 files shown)


athas:
Some of these packages are not in the `uv` config, and having them in the `shell.nix` means you cannot them outside of `uv run` (or whatever that command is). This is a minor issue that I could personally live with.

More seriously, I also recall there are problems with `uv` not working well (or at all) on NixOS, since it has assumptions about binary compatibility. I will give that a try.

athas:
```
$ uv run python/gradbench/gradbench/evals/ba/run.py
error: Querying Python at `/home/athas/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11` failed with exit status exit status: 127

[stderr]
Could not start dynamically linked executable: /home/athas/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```

I wonder if we can just change the Nix configuration to provide the Python version that `uv` expects.

athas:
Yes, if we explicitly make the expected version of Python (3.11) available via Nix (by adding `pkgs.python311`), then `uv` just uses that instead of downloading something. I still think there may be trouble with all the other random binary blobs it downloads, but we'll see.

*Generally* I will say that mixing Nix (in particular NixOS) and other package managers is usually dubious, *especially* those that have the misdesign of being based on precompiled binaries, but we'll see.

samestep:
> Some of these packages are not in the `uv` config, and having them in the `shell.nix` means you cannot them outside of `uv run` (or whatever that command is). This is a minor issue that I could personally live with.

Not sure I understand this; you're saying that _having_ these packages in `shell.nix` means you _can't_ use them outside of `uv run`? How does that work? And if that's the case, then doesn't that mean it makes even more sense to remove them from `shell.nix`? (I'm guessing there's simply some typo in your paragraph which is causing my confusion.)

> ```
> $ uv run python/gradbench/gradbench/evals/ba/run.py
> error: Querying Python at `/home/athas/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11` failed with exit status exit status: 127
> 
> [stderr]
> Could not start dynamically linked executable: /home/athas/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/bin/python3.11
> NixOS cannot run dynamically linked executables intended for generic
> linux environments out of the box. For more information, see:
> https://nix.dev/permalink/stub-ld
> ```

Ah gotcha, thanks. For some reason, though, I don't see that same error on my own NixOS machine?

> Yes, if we explicitly make the expected version of Python (3.11) available via Nix (by adding `pkgs.python311`), then `uv` just uses that instead of downloading something.

OK sweet! Let's just do that, then?

> I still think there may be trouble with all the other random binary blobs it downloads, but we'll see.

What other binary blobs does it download?

> _Generally_ I will say that mixing Nix (in particular NixOS) and other package managers is usually dubious, _especially_ those that have the misdesign of being based on precompiled binaries, but we'll see.

Hmm what do you mean? Precompiled binaries aside, isn't this already exactly what we do with `cargo`? Seems to work just fine. And this is just for local development; if we wanted to actually package any of this up using Nix, we wouldn't use either `cargo` _or_ uv, right? In that case we'd use [Naersk](https://github.com/nix-community/naersk) or [`uv2nix`](https://github.com/pyproject-nix/uv2nix) instead. But that's not the situation we're actually in, correct?

samestep:
@athas I've added `pkgs.python311` to `shell.nix`. Does this fix the error you were seeing on your NixOS machine?

athas (approved):

PR #397, opened 2025-03-22
New evals: ode, llsq, det

athas:
These are from cmpad. I have ported all of them, although not yet for all tools.

Before anyone asks, I don't know why it is called `an_ode` and not just `ode`, but I felt it was better to keep the original names.

There is a fair amount of additional mathematical information about the benchmarks in the cmpad documentation. I have opted not to copy it, for various reasons:

  * I am not confident in my ability to transcribe it without introducing errors.

  * GitHub's Markdown-math support is not particularly good.

  * The cmpad documentation is still available.

And honestly, the cmpad problems are not very exotic, so implementers can find suitable information in many places.


samestep (approved):
Amazing! No notes.

> GitHub's Markdown-math support is not particularly good.

Fair enough; but hey, looks like it may have gotten slightly better recently? https://github.com/microsoft/vscode/issues/208430#issuecomment-2762344110

PR #399, opened 2025-03-22
Add all dependencies to LD_LIBRARY_PATH

athas:
This helps some programs (mostly or exclusively Python libraries in our case) that search LD_LIBRARY_PATH using bespoke bechanisms that the usual Nix automation has not patched.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -19,7 +19,7 @@ let
   adept = pkgs.callPackage ./nix/adept.nix { };
   codipack = pkgs.callPackage ./nix/codipack.nix { };
   GRADBENCH_PATH = builtins.getEnv "PWD";
-in pkgs.stdenv.mkDerivation {
+in pkgs.stdenv.mkDerivation rec {
   name = "gradbench";
   buildInputs = [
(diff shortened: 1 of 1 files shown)


athas:
This is much more effective than I expected, e.g. this works fine:

```
$ ./gradbench run --eval 'uv run python/gradbench/gradbench/evals/gmm/run.py' --tool 'uv run python/gradbench/gradbench/tools/tensorflow/run.py'
```

I cannot explain why this works.

samestep (approved):
> This is much more effective than I expected, e.g. this works fine:
> 
> ```
> $ ./gradbench run --eval 'uv run python/gradbench/gradbench/evals/gmm/run.py' --tool 'uv run python/gradbench/gradbench/tools/tensorflow/run.py'
> ```
> 
> I cannot explain why this works.

Huh, that actually fails on my machine with [a ton of CUDA errors](https://github.com/user-attachments/files/19406233/errors.txt). But I don't really care about running TensorFlow outside of Docker on my NixOS machine, so this is fine with me!

PR #401, opened 2025-03-24
intermediary: kill tool on timeout

### LINKED ISSUE #374: Timeout takes a long time to actually stop
For example, the command I used in #373 took about a minute after the `⧖` was printed before I actually got my next terminal prompt. It should instead exit pretty much immediately, at least in under a second.

I'm guessing this will end up sharing a solution with #364.

athas:
This kills the tool process when the intermediary fails with a timeout.

Fixes #374, but will in most cases not actually kill all subprocesses spawned by the tool, in case they are in a sub-process group.

### DIFF
--- a/crates/gradbench/src/intermediary.rs
+++ b/crates/gradbench/src/intermediary.rs
@@ -405,7 +405,7 @@ pub fn run(
         }
     }
     let start = Instant::now();
-    Intermediary {
+    let outcome = Intermediary {
         outcome,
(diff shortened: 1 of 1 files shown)


athas:
Also the Rust here is crummy, feel free to rewrite it.

samestep (approved):
Sweet, thanks for the fix!

> Also the Rust here is crummy, feel free to rewrite it.

Nah I think it's fine. I'm just gonna fix the Clippy lint.

PR #405, opened 2025-03-26
Add logo with random gradient to gradben.ch

athas:


samestep:
what the hell, I've never seen this Copilot PR review thing before

athas:
GitHub suggested it, and who am I to not click a button?

samestep:
Ah OK, I was just worried it appeared automatically with no one triggering it. If you did it by clicking a button then that's fine.

Unfortunately most of its suggestions are not particularly good; I'm going to make some of my own improvements before merging this, though.

athas:
I like the suggestion that would hide a bug but not actually avoid it. A program is correct precisely when it does not throw any exceptions, right?

samestep (changes requested):
I still plan to tweak this so we don't need to `useRef`, but apart from that, the only other issue is that the formatting is not great on mobile. Granted, the website already isn't great on mobile (should work on that), but this makes it a bit worse; can we tweak the layout to make it look a bit better?

| before | after |
| - | - |
| before | after |

athas:
It now looks better on narrow screens.

samestep:
OK now we programmatically generate the logo on the website as of dffda29d38446d661cc48eb3f86561ab03fdac9d; tomorrow morning I'll see if I can use that same `Logo.tsx` to programmatically generate the logo we use for `README.md`, letting us delete `logo.svg` from Git.

samestep (approved):
OK, I think I'm finished. @athas does this look good to you? Sorry for taking so long.

athas:
How do I run the website locally? Did the command change?

athas:
I [RTFM](https://github.com/gradbench/gradbench/blob/main/CONTRIBUTING.md#website).

athas:
Very shiny!

PR #407, opened 2025-03-27
Compute stats for particle and saddle

athas:
Plotting will take more work, but this at least makes the grid look better.

### DIFF
--- a/crates/gradbench/src/stats.rs
+++ b/crates/gradbench/src/stats.rs
@@ -198,11 +198,81 @@ impl<R: BufRead, F: CreateFile> Scorer<R, F> for ScorerClassic {
     }
 }
 
+#[derive(Serialize)]
+struct ScorerParticleSaddle {
+    /// The average duration for each tool (outer keys) and function (inner keys).
+    tools: BTreeMap<String, IndexMap<Rc<str>, Duration>>,
+}
+
+impl ScorerParticleSaddle {
+    fn new() -> Self {
+        Self {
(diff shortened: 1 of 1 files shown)


crates/gradbench/src/stats.rs
```diff
@@ -198,11 +198,89 @@ impl<R: BufRead, F: CreateFile> Scorer<R, F> for ScorerClassic {
...
+///
+/// The point of these evals is to investigate the consequences of
+/// different implementation choices. In a real setting, one would
+/// presumably pick the best one, so the score is reported as the
+/// minimum runtime achieved.
+#[derive(Serialize)]
+struct ScorerParticleSaddle {
+    /// The average duration for each tool (outer keys) and function (inner keys).
+    tools: BTreeMap<String, IndexMap<Rc<str>, Duration>>,
+}
```
samestep, line 213: Thanks for the nice documentation! Since (like `ScorerClassic`) this scorer is based around an idea rather than only being suitable for `particle` and `saddle`, can you change the name to be more general, to be consistent with how `ScorerClassic` is already named?

samestep (changes requested):
Looks good, thanks! Just had one minor request.

Perhaps after we merge this we can take a look at the implementations of `ScorerClassic` and this new scorer to see if there are any opportunities to deduplicate and make the code nicer, but I don't think that's necessary in this PR.

samestep (approved):

PR #412, opened 2025-03-29
Add Commentary

athas:
The idea behind the Commentary is to have a place for non-normative information about evals and tools. We can imagine the Commentary helps in the following cases:

* You are writing a paper about some new AD tool, and you need to explain why some of the existing implementations in GradBench are particularly slow or fast.

* You want to add a new tool to GradBench, and you are looking for advice on which evals to start with.

### DIFF
--- a/tools/futhark/README.md
+++ b/tools/futhark/README.md
@@ -10,4 +10,32 @@ To run this outside Docker, you'll first need to install Futhark and run the fol
 futhark pkg sync
 ```
 
+## Commentary
+
+Most of the implementations are clean and perform well, relative to
+the expectations of a pure functional language. The implementations
+are always (when algorithmically possible) written in a parallel
+style, even though the programs are run sequentially by default, and
+sometimes the parallel style is slower in that case.
+
+* [det.fut][]: Futhark does not support parallelism, so this is
+  implemented in a rather different way that is less efficient but
+  data parallel, based on computing and applying permutations. The
+  amount (and form) of floating-point work is the same, so this is in
+  spirit still the same (inefficient) algorithm as specified by the
+  eval.
+
+* [saddle.fut][] and [particle.fut][] make use of [solver.fut][],
+  which is a rather simple implementation of Gradient Descent. Their
+  performance is slightly hampered by representing two-dimensional
+  points as two-element arrays, which is [not particularly efficient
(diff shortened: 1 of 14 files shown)


athas:
I'll add more than this eventually, but I would like to know if the form is useful. It has been my experience in other projects that it is useful to have an explicitly unstructured place to put documentation when you have something that *the world must know*, but it doesn't fit naturally within the existing structure.

tools/futhark/README.md
```diff
@@ -10,4 +10,36 @@ To run this outside Docker, you'll first need to install Futhark and run the fol
...
 
+## Commentary
+
+Most of the implementations are clean and perform well, relative to
+the expectations of a pure functional language. The implementations
+are always (when algorithmically possible) written in a parallel
+style, even though the programs are run sequentially by default, and
+sometimes the parallel style is slower in that case.
+
+- [det.fut][]: Futhark does not support parallelism, so this is
```
samestep, line 21: I assume this is a typo?

```suggestion
- [det.fut][]: Futhark does not support recursion, so this is
```

tools/futhark/README.md
```diff
@@ -10,4 +10,36 @@ To run this outside Docker, you'll first need to install Futhark and run the fol
...
+
+- [saddle.fut][] and [particle.fut][] make use of [solver.fut][],
+  which is a rather simple implementation of Gradient Descent. Their
+  performance is slightly hampered by representing two-dimensional
+  points as two-element arrays, which is [not particularly efficient
+  in Futhark][]. The implementation is otherwise very clean, and works
+  by defining a generic function that is then instantiated by the four
+  different combinations of differential operators.
+
+- [ba.fut][]: Just as much time is spent packing the sparse Jacobian
```
samestep, line 36: Broken link, can you fix?

samestep (approved):
This is excellent and extremely useful, thanks so much for writing these up. The only edits I made in 26d69ce117adeb6b9e8c07d07e23d670fb913a6c were to replace relative Markdown links with absolute ones, since e.g. the link to the PyTorch tool in `evals/llsq/README.md` was broken since the relative links are easy to mess up. Other than that, I just had a couple comments on typos in `tools/futhark/README.md`.

PR #461, opened 2025-03-31
intermediary: set 'docker build' stdout to /dev/null

athas:
Only done on the quiet path. This is critical for Podman-based Dockers, which print diagnostics to stdout. Otherwise, 'gradbench repo tool' will not work, as the diagnostics interfere with the protocol.

### DIFF
--- a/crates/gradbench/src/main.rs
+++ b/crates/gradbench/src/main.rs
@@ -302,6 +302,10 @@ enum Verbosity {
 fn docker_build_quiet(color: Color, mut cmd: Command) -> anyhow::Result<ExitStatus> {
     let mut child = cmd
         .arg("--progress=plain")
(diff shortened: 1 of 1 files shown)


crates/gradbench/src/main.rs
samestep, line 1: Seems reasonable to me. Can you add a comment explaining that this is specific to Podman, ideally with a relevant link?

samestep:
Also just to clarify, is this a fix for #465, or unrelated?

athas:
It is unrelated.

crates/gradbench/src/main.rs
athas: I have added a comment. I can't find any normative link that describes this Podman behaviour (but it honestly makes sense from Podman's perspective).

samestep (approved):

Issue #338, opened 2025-03-05
Strategy for multiple runs is inflexible


samestep:
In #151 we added a `--runs` CLI argument to the GMM eval, which was a simple integer that got included as a `"runs"` field in the `"input"` for `"evaluate"` messages, and dictated the number of times to repeat running the function. Then in #162 we extended that to the other evals from ADBench and defined a `@wrap.multiple_runs` Python decorator to make it easy to implement this for Python tools. In #166 we also implemented this for several C++ tools.

Then in #282 we decided that this was not flexible enough, so we replaced `--runs` with `--min-runs` and `--min-seconds`, which get translated to `"min_runs"` and `"min_seconds"` in the `"input"`. It's up to each individual tool to implement this correctly, and currently, some still don't; for instance, the SciLean GMM implementation in #224 just ignores these fields, as mentioned in #329.

Going forward, we may want to make our stopping strategy more sophisticated, perhaps by dynamically estimating statistical uncertainty as suggested by @athas [on Discord](https://discord.com/channels/1242864568850579487/1267596255886970952/1346887647951589488). If we do this, it would be unwise to duplicate this code change across all the languages represented by different tools. Instead, I propose that we remove `"min_runs"` and `"min_seconds"` from the `"input"` sent to the tool, and handle this at the protocol level.

Specifically, I propose that we add a new `"kind": "repeat"` message, with the following rules:

1. Like `"kind": "analysis"`, there will be an `"of"` field containing the ID of a previous `"kind": "evaluate"` message to repeat.
2. To simplify the caching logic required by tools, there must have been _no_ other `"kind": "evaluate"` messages since the one that is to be repeated.
3. The tool should re-run the function from that previous message with the same input, and return a response with all the same fields _except_ for the `"output"`.

Then it's the tool's job to cache the most recent input, and it's the eval's job to decide when to stop telling the tool to repeat. This will also require changes to the `gradbench` CLI to handle this new way of recording timing measurements.

athas:
This is a somewhat complicated issue. Here is an itemized list of opinions and observations, for ease of rebuttal.

1. It would be ideal if evals could control their measurement logic independently of tools.

2. The current measurement logic is pretty good, exceeding the quality of most other benchmark suites, and so it is not worth to put in a big effort to improve it, although a small effort could be justified.

3. With the `min_seconds`/`min_steps` protocol, the logic is duplicated across tools, but this is not a big problem, as it is very easy to implement. I have done so many times already, in various languages.

[...]

PR #464, opened 2025-04-03
Refactor Julia setup

vchuravy:
Inspired by the Python setup and the C++ setup it would make sense to collect the Julia benchmark in a package,
to avoid code duplication between Enzyme and Zygote as an example.


athas:
This looks good. It also struck me last I looked at the Julia setup that it didn't seem right.

julia/Gradbench/src/benchmarks/ode.jl
```diff
@@ -0,0 +1,64 @@
...
+#
+# https://github.com/bradbell/cmpad/blob/e375e0606f9b6f6769ea4ce0a57a00463a090539/cpp/include/cmpad/algo/runge_kutta.hpp
+#
+# originally by Bradley M. Bell <bradbell@seanet.com>, and used here
+# under the terms of the EPL-2.0 or GPL-2.0-or-later.
+#
+# The implementation in cmpad is factored into a generic Runge-Kutta
+# module and an instantiation for the specific function under
+# consideration. In this implementation, this is all inlined for
+# implementation simplicity. This is generally written in a more
```
athas, line 11: Surely this remark about C++ is not pertinent to Julia?

vchuravy:
I had a question about the JSON protocol. The `timings` response can contain multiple answers? How is that meant to be used? Should I measure N times?

athas:
Yes, see the `min_seconds` and `min_runs` specification here: https://github.com/gradbench/gradbench/blob/main/CONTRIBUTING.md#specification

athas:
See also the C++ timing code here: https://github.com/gradbench/gradbench/blob/543ff10b5f5e8e314220dc7e47f6e05d171e2d57/cpp/gradbench/main.hpp#L49-L56 It is not a complicated algorithm.

julia/GradBench/.gitignore
samestep, line 1: @vchuravy why is `Manifest.toml` in `.gitignore` here? My understanding is that it's a lockfile analogous to our existing `Cargo.lock`, `uv.lock`, and `bun.lock` files, correct?

tools/zygote/Dockerfile
samestep, line 1: Thanks for updating the Dockerfile! Apologies for not having this written explicitly in `CONTRIBUTING.md`, but our convention is to have `WORKDIR /gradbench`. Does it work to continue using that same convention here?

athas: My understanding, based on talking to Valentin at EuroAD, is that this path makes it more convenient to also run the tool outside of Docker (something about the manifest being sensitive to relative paths).

samestep: OK! In that case @vchuravy could you add a comment explaining why this structure makes running outside Docker easier, so we don't accidentally change it in the future?

julia/GradBench/src/GradBench.jl
```diff
@@ -0,0 +1,73 @@
...
+    @assert min_runs > 0
+
+    # TODO: Prepare (parse JSON?)
+    # TODO: pre-allocate output?
+    timings = Any[]
+
+    # Prepare
+    # - run once to compile
+    ret, t = measure(func, arg)
+    push!(timings, Dict("name" => "prepare", "nanoseconds" => t))
```
samestep, line 38: One other request: please don't run once in the prepare stage to pre-compile. One goal of GradBench is to capture the fact that some tools pre-compile and other tools use JIT compilation, and we use multiple runs to gather the data for that. For instance, JAX also uses JIT compilation, and CoDiPack analogously does more work on the first run than others. The `prepare` step should be for tasks other than actually running the program.

athas: The precompilation overhead of Julia is so high that you would get zero data on the first run on anything else than the compilation time, and then it would be invisible on subsequent runs. This is not a particularly useful way of presenting this information. I think precompilation as part of `"prepare"` or even `"define"` is fine, as long as the time is reported.

`cppad` does precompilation (of a sort) in `prepare`, and so does `adol-c`, so it is not new to GradBench. I'm actually not sure how Jax gets away with it.

samestep: > The compilation overhead of Julia is so high that you would get zero data on the first run on anything else than the compilation time, and then it would be invisible on subsequent runs. This is not a particularly useful way of presenting this information.

Can you clarify what you mean by this? I don't understand exactly what you mean.

> I think precompilation as part of `"prepare"` or even `"define"` is fine, as long as the time is reported.
>
> `cppad` does precompilation (of a sort) in `prepare`, and so does `adol-c`, so it is not new to GradBench. I'm actually not sure how Jax gets away with it.

Do either of those involve actually running the program _on the data_, or are they independent of the data?

athas: Looking at the Jax numbers, I think Jax just has a (relatively) fast compiler (for the two evals we actually implement with Jax), so we have not noticed that the results are a bit skewed.

athas: It's actually a bit of an all-or-nothing situation: *if* the compiler runs in less than a second, then you get up to that second to perform as many runs of the (presumably tiny) smallest workload to bring down your average runtime. If the compiler takes at least a second, then that's it, your time's over and that (single) run gets to be your measurement.

samestep: My position is that for tools whose first run is significantly slower than further runs, we should handle that in the way we tell the tool how many times it needs to run, and in the way we compute and visualize statistics, not by having the tool account for it itself. The tool shouldn't get to decide what is or is not a relevant performance statistic.

samestep: Regarding the one-second cutoff: right, that's part of the reason I still somewhat think #338 might be a good idea, so we can iterate on our strategy for multiple runs. But in this case that's probably overkill; seems like the solution is just to set `min_runs` to 2 instead of 1.

samestep: Or perhaps we should use a more sophisticated strategy that accounts for the fact that many tools only have the _first_ run be slower, and so ignore that when telling them how many times they should keep going. But again that goes more into the realm of #338.

athas: > Can you clarify what you mean by this? I don't understand exactly what you mean.

In Julia, the function is compiled when it is first run, but compilation does not *depend* on the input, and is not influenced by it. Hence the first run will be slow (because the compiler is called for the first time), but subsequent runs will be fast (the function has been compiled, even though the input changes). That is just how Julia works. The compilation will usually be *very* slow, because it also touches a bunch of the stdlib for the first time. ADBench's Julia implementations address this by running the benchmark code on a tiny "synthetic input" when it is imported.

> Do either of those involve actually running the program _on the data_, or are they independent of the data?

Yes, they construct the "tape", which is then evaluated multiple times. This does reflect a somewhat common use case for these tools, where the tape is used multiple times with slightly different inputs (but of the same "shape"). This is documented as part of the ADOL-C README: https://github.com/gradbench/gradbench/tree/main/tools/adol-c#readme I intentionally kept it that way partly because it reflects ADOL-C usage, and partially to eventually start a debate on the issue.

For CppAD the issue is somewhat more severe: while ADOL-C benefits from precomputing the tape, it is not a complete deal-breaker. However, CppAD has *bad* performance without the use of "tape optimization", and it is clear that this is somehow part of how CppAD is supposed to be used. That does not mean we should let it get away with it, however.

I think the idea is that although this is "precompilation" on specific input, it is not "specialization" - you can actually change the primal values after the tape has been constructed and rerun it for new input, as long as the "structure" is the same. But I am certain that my interpretation is correct. (You can certainly do this for new *seed values* however, which may be how these libraries are actually supposed to be used - but that is not currently something we do in GradBench.)

athas: A very easy solution is to always ask for more than one run, but then always discard the first one. That's actually how I did it when I implemented `futhark bench`, *specifically* to avoid noise related to GPU JIT compilation of kernels.

samestep: Thanks for the added context. Given that, I'll clarify my position:

1. If a tool has work that it needs to do before it can run fast, and that work is independent of the input, then it should do it when it receives the `define` message.
2. If a tool has work that it needs to do before it can run fast, but it can only do that work _while running on a real input for the first time_, then it should report that the exact same way as the rest of its timings for running the function fast later.
3. If a tool has work that it needs to do that is purely an artifact of GradBench's JSON-based protocol design, it can either not report it or report it as a separate timing, e.g. called `prepare`.

> A very easy solution is to always ask for more than one run, but then always discard the first one. That's actually how I did it when I implemented `futhark bench`, _specifically_ to avoid noise related to GPU JIT compilation of kernels.

Sure, this sounds reasonable to me: all I'm asking is that it's a decision we make in the centralized infrastructure (so that we then have the freedom to collect other statistics and make other visualizations, e.g. about startup time), not in the tool.

julia/GradBench/.gitignore
samestep: Makes sense, thanks for clarifying!

julia/GradBench/src/GradBench.jl
athas: I don't have a problem with any of this.

Is the most practical solution to set `min_runs` to 2 and then update intermediary/plotting/etc to discard the first run when computing and plotting averages?

The main downside is that we have some very large workloads where we currently do only a single run, and doing two would... well, double the time taken.

samestep: Yeah I think that's fine, right? And then if that causes those tools to timeout on more evals then sobeit.

samestep: Yeah, for Julia, precompilation seems like the best approach.

athas:
I fixed a runtime error and it seems to work now. Should we merge it and continue work in future PRs?

samestep (approved):
Thanks @vchuravy for the refactor!

athas:
Excellent foundation. With 14 AD libraries, at least one of them ought to be usable for GradBench. I will try to poet the Zygote code from ADBench, but it would be very appealing to go through DifferentiationInterface.jl.

vchuravy:
> but it would be very appealing to go through DifferentiationInterface.jl.

So the biggest variance will be mutation support or not. Porting the ODE example with Zygote immediately fails.
So we will have some tension between faithful reproduction of the benchmarks, idiomatic code for Julia, and idiomatic code for the benchmark AD framework.

The problem is what we are introducing two more axis of parameter exploration. The ODE benchmark could have an out-of-place implementation that Zygote could support, and then the AD benchmark might have some choices as well. 
Enzyme alone has "Reverse" / "Forward"  and "combined primal with adjoint" and "separate primal and adjoint".

DifferentiationInterface adds N different backends. 

Maybe evals.txt could provide settings for the benchmarks?

```
hello
ode
ode inplace=true
ode backend=enzyme
ode inplace=true forward=true
```

athas:
> So we will have some tension between faithful reproduction of the benchmarks, idiomatic code for Julia, and idiomatic code for the benchmark AD framework.

I think you listed these in increasing order of priority. I view GradBench as an experiment answering the question "if a competent user wanted to solve this problem with this AD framework, what would they do and how well would it work?" Tools (at least some of them) have a *Commentary* section in their READMEs where we can document if a tool really has to turn the problem specification inside out to make it work.

For example, it is *not* a requirement that all of the Julia implementations use the exact same underlying primal code, even if/when we switch to DifferentiationInterface.jl - although the hope is of course that significant sharing will happen. I'm not sure exactly how we would wrangle it, but we might end up with different versions of the code, and a tool can pick whatever fits best (or even have a completely bespoke version in the extreme case). As long as we clearly document what we have done (and why), I think that will show something interesting about the different tools (this is the "comparison, not competition" angle).

samestep:
@vchuravy to be a bit more concrete: we'd still want every separate Julia autodiff tool to go in its own subdirectory under `tools/`, but what Troels is saying is that it might be wise to use DifferentiationInterface to deduplicate code across most of the Julia tools.

As you mentioned, there are important differences between tools! So for instance if Zygote is unique in not supporting mutation, maybe it'd share less code with the other Julia tools. And if Enzyme is unique in providing "combined primal with adjoint" and "separate primal and adjoint" then maybe it'd have more code specific to it.

You can see this in the `python/` directory in this repo. Many tools use parts of our Python codebase, but no single tool uses _all_ the Python code. It's up to the tool's `Dockerfile` to point to the parts of the codebase that make the most sense for that particular tool.

athas:
Interestingly, the Python tools share *no* primal code - it's all reimplemented from scratch for every tool. Only the IO and measurement boilerplate is shared. This is not due to principles, but simply because there is not any useful shared API between e.g. PyTorch, TensorFlow, and Jax (except to the extent they emulate Numpy, perhaps).

samestep:
Right yes; perhaps the C++ codebase would have been a better example, then.

PR #467, opened 2025-04-03
Temporarily disable `lstm` benchmark for SciLean

lecopivo:
I can't reproduce the issue with `lstm` in a separate executable. Right now, I have no idea how to debug this. So this PR temporarily disables `lstm` for SciLean

### DIFF
--- a/tools/scilean/Main.lean
+++ b/tools/scilean/Main.lean
@@ -15,7 +15,7 @@ def resolve (module : String)
   | "hello" => some hello
   | "gmm" => some gmm
   | "kmeans" => some kmeans
-  | "lstm" => some lstm
+  -- | "lstm" => some lstm
   | "llsq" => some llsq
   | _ => none
 
--- a/tools/scilean/evals.txt
+++ b/tools/scilean/evals.txt
@@ -2,4 +2,3 @@ gmm
 hello
 kmeans
 llsq
-lstm


athas:
Is this just a draft, or do you want us to merge it?

lecopivo:
Yes merge it, not sure when I get to fix this

lecopivo:
I'm using OpenBLAS implementation of BLAS it is says it uses multi-threading. 

But running
```
./gradbench run --eval "./gradbench repo eval lstm" --tool "export OPENBLAS_NUM_THREADS=1; ./gradbench repo tool scilean"
```
still fails sometimes with the same error.

samestep:
Huh, does that environment variable successfully pass through Docker?

lecopivo:
Ha that is why I wrote down the exact command I'm running as I have no clue :)

Notice that I'm setting it inside of the `tool` argument not on the top level command. But I have no idea if it actually works

samestep:
Yes, you can! The `./gradbench repo tool scilean` command just builds a Docker image using `tools/scilean/Dockerfile`, and then runs that. So you can just run the same command as your Docker `ENTRYPOINT` instead.

lecopivo:
Bazinga! This works
```
./gradbench run --eval "./gradbench repo eval lstm" --tool ./tools/scilean/.lake/build/bin/gradbench
```
Mentioning this with an explicit example [here](https://github.com/gradbench/gradbench/blob/main/CONTRIBUTING.md#docker) would be nice.

samestep:
> Mentioning this with an explicit example [here](https://github.com/gradbench/gradbench/blob/main/CONTRIBUTING.md#docker) would be nice.

We do already mention running outside Docker [in the Python section](https://github.com/gradbench/gradbench/blob/a796123d5b59eff5fd9497a9532f902958fb59a5/CONTRIBUTING.md#python), but I agree that we should mention it further up since it's not Python-specific.

athas:
The specific command always depends on the tool. We should ask the maintainer of the `scilean` tool to document this in their README.

lecopivo:
That is true but every example in the readme passes a string into that argument. It didn't even occur to me that I could pass the path to the executable.

samestep:
They're all strings, whether or not you use quotes; that's just how shells work. The only difference is that you can choose not to use quotes if your string has no spaces in it, which is true in this case because you're not passing any additional arguments to `./tools/scilean/.lake/build/bin/gradbench.

lecopivo:
Haha I just revealed my complete CS fundamentals ignorance :) ... interesting, I can type `"cd" /usr`

samestep:
Lol I wish we didn't have to know all these random facts and syntaxes etc... but anyways this is a good point, probably a good idea to spell things out more explicitly by giving examples like this without quotes, since not everyone is going to have obsessively read through a manual on Bash.

athas:
I think any computer scientist would be sad if shell word splitting syntax was now considered "CS fundamentals". It's just implementation details from the 70s. Anyway, I added this to `tools/scilean/Dockerfile`:

```
ENV OPENBLAS_NUM_THREADS=1
```

But it does not fix the issue.

PR #470, opened 2025-04-04
Rewrite ht-adol-c to use forward mode

athas:
Together with vectorises AD, this provides about 2.5x speedup over the old implementation. It is almost as fast as CoDiPack.


athas:
@TimSiebert1 it worked fine, but maybe you want to check out my code anyway.

samestep (approved):
Sweet!

PR #472, opened 2025-04-04
cppad: don't do input-dependent computation in constructor

athas:


samestep (approved):
Thanks for making this more consistent with the design discussion from #464!

athas:
I haven't gotten to all the evals yet. One important impact of this change is that it removes all use of the CppAD `optimize` method on tapes. When we count its runtime, then I have not yet seen a case where the benefit outweights its expense.

samestep:
@athas can you clarify what you mean? Are you saying that this is not just making the computation be reflected in the `"evaluate"` timings, but also turning off optimizations?

athas:
Yes, because those optimizations are extremely costly. They just make the whole thing slower. They may be worth it for evals that run the tape multiple times (we'll see if I can make it do something useful for `ht` and `ba`). Otherwise, my hunch is that they are intended for computation of huge dense Jacobians.

athas:
Oh, actually they look like they are worth it for `kmeans`. Probably related to the sparsity of the Hessian. It's still pretty slow overall.

PR #395, opened 2025-03-21
Revamp Python build infrastructure

samestep:
Resolves #236, resolves #320. As a followup on #394, I've updated the Nixpkgs version because this new part of `pyproject.toml` depends on [a feature](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups) introduced in [uv 0.6.8](https://github.com/astral-sh/uv/blob/0.6.8/CHANGELOG.md#enhancements):

```toml
[tool.uv]
default-groups = "all"
```

While I was at it, I also cut out a ton of Python boilerplate for our C++ tools.

PR #473, opened 2025-04-05
Remove run_ prefix from C++ implementations

athas:
There is no need for this anymore after Sam cleaned up the infrastructure in #395.

I also added some more documentation for the C++ infrastructure while I was at it.


athas:
Most of the non-C++ tools (Haskell, OCaml, SciLean) already do it that way, although with different directory names. I suppose it would not be difficult to do a unified thing for all of the C++ tools. There is also a lot of redundancy in the `Makefile`s - I somewhat expect that if someone with a true passion for C++ eventually shows up to work on GradBench, one of the first things they will do is replace the build system with cmake or whatever is fashionable nowadays.

But these issues mostly affect people who want to run the C++ code outside of Docker, so it's pretty far down on my priority list.

Issue #333, opened 2025-03-03
Markdown is not linted


samestep:
It's annoying and error-prone to manually look for mistakes like the one fixed by #332. Perhaps we should use something like [`markdownlint`](https://github.com/DavidAnson/markdownlint).

PR #474, opened 2025-04-05
Begin work on multi-threading

### LINKED ISSUE #303: Tools only run on one core
In #301 we used [`taskset`](https://man7.org/linux/man-pages/man1/taskset.1.html) to limit JAX to a single core, for a more accurate comparison against all the other tools running on one core. However, this shouldn't be the only configuration we consider: parallelism is a key feature of many tools, and it'd be good to measure that in CI. Perhaps we should require that all tools in this repo support a CLI argument called `--cores` that takes a positive integer and limits that tool to use a maximum of that many cores.

Here's some data from [the GitHub docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories):

- `ubuntu-24.04` is x86 and has 4 cores
- `macos-14` is ARM and has 3 cores

The latter only matters once we address #302. As mentioned in that issue, we'd want to add a radio button to the website to select the number of cores in the data being shown. But while it doesn't really make sense to compare x86 data against ARM data, it does make sense to compare data for different numbers of cores on the same architecture. In #293 we discussed clicking on a tool name to toggle showing or hiding it; another possibility could be that, when only one tool is selected, the website shows the data for all possible values of `--cores` simultaneously, similar to how it currently shows the data for multiple tools simultaneously as different colors. That would allow the user to directly see how well each tool takes advantage of multiple cores.

athas:
This PR begins the work of multi-threading the GradBench implementations. There are three main concerns:

1. Having multi-threaded implementations.
2. How to control sequential versus parallel execution, e.g., whether the parallel variants should be separate tools.
3. How/if/when this should be run in CI, and what should be done with the results.

For now I am focusing entirely on 1. In particular, I am multi-threading those C++ implementations where this can be done with OpenMP, without duplicating any code, and without regressing the performance of the single-threaded case.

It is not clear how to handle the cases where a problem can be solved in both a sequential and a parallel way, and the parallel way running as a single thread is less efficient than a specialised single-threaded implementation. I have my own opinions on how this should be resolved, but for now I have been able to avoid any such cases.

To actually run these implementations, I extended `cpp.py` such that it accepts an option `--multithreaded`. If set, this enables a flag in the Makefile (currently only for `manual`) that enables OpenMP.

Some of the evals are completely sequential and cannot be parallelised. `particle` and `saddle` are two such cases, and in practice also `det` (unless you write it in the convoluted way I did for Futhark, but it is not a good idea). This is fine. GradBench is not specifically concerned with parallel execution.

For many of our workloads, parallel execution is slower than sequential execution. This is *also* fine. GradBench is not supposed to be a collection of the best way to solve some (primal!) problem for all workloads. The purpose of the multi-threaded implementations is to investigate how this impact the performance of AD'd code, and this is largely orthogonal to the issue of whether multi-threading is a good idea for some workload in particular.

Some of our implementations are already written in a parallel style, and can be switched to that mode easily enough. This is certainly the case for Futhark, PyTorch, TensorFLow, and Jax. I am mostly focusing on the C++ implementations for now. I am particularly concerned with how the tape recorders will handle multi-threading - I think many of them are unable to do the recording in parallel, but may be able to perform multiple passes of the tape using multiple threads.

Closes #303.


athas:
Looks like Enzyme does not support OpenMP reduction clauses: https://github.com/EnzymeAD/Enzyme/discussions/1251

This puts rather a damper on my hope that this could be a noninvasive change, as it requires a little more work to write these programs without reductions.

samestep:
Thanks for working on this! Everything you've written in the PR description here sounds pretty reasonable to me. I don't current have strong opinions about any of this: I laid out some preliminary thoughts about your third bullet point in #303, but for the rest, I haven't thought much yet.

athas:
I am considering whether we should merge this even when incomplete. It does not make the sequential case worse, and it will take a very long time to multithread everything as well as it can be. In some cases we are held back by things like Enzyme bugs, which we are unlikely to fix by ourselves.

samestep:
@athas either way sounds reasonable; I'd say that judgment call is up to you. I looked over this PR and it seems pretty good to me, although I think we still need to discuss the high-level design a bit. Specifically, in the past we discussed having a `--cores` CLI arg that tells the tool how many CPU cores it's allowed to use, but I see here that it's collapsed into just a binary `--multithread` arg. Is there a particular reason for doing that rather than the more granular `--cores` interface?

athas:
Whether to use multiple threads or not is a binary algorithmic issue that tools need to be specifically told about so they can take precautions, like using specific code paths. *How many* cores they get to use is an issue that is better handled with `taskset`. It is also likely that many tools don't actually provide any good way of controlling core count - Jax doesn't even provide a builtin way of forcing single-threaded execution.

samestep:
For sure, I totally agree; but `taskset` still needs to be done inside of the Docker container, right? That's what I'm saying: from the user perspective, it seems like the interface should be `--cores`, and then yeah it seems totally reasonable to internally translate values of 2 or above for `--cores` to a `--multithread` parameter when compiling etc. Or am I misunderstanding?

samestep:
Sorry, I think I was being a bit unclear: I didn't mean to put `taskset` literally into the `Dockerfile` as we currently do for JAX; I was always under the impression that that was just a temporary proof of concept.

What I meant was that the CLI arg being passed to the tool should be `--cores`, an integer. Then it's up to the tool to decide what to do with this; probably what we'll end up doing is to have a single script that pretty mosts tool use, which does a couple things:

1. Call `taskset`.
2. Pass a script to the "actual" tool implementation.
3. Pass a `--multithread` argument to that script, and/or set a `MULTITHREAD` environment variable.

Does this sound reasonable? Or am I still missing something?

samestep:
OK! Sounds like we're in agreement. Just to clarify, to make sure I understand the proposed design:

- Every tool accepts an integer `--cores` argument, which will probably just get passed to `taskset`.
- Some tools may also choose to accept a boolean `--multithread` or `--multithreaded` argument (I don't care which of those two names we use, but let's be consistent), which can default to being determined by `--cores` if not provided.

Is this what you have in mind?

athas:
> Every tool accepts an integer --cores argument, which will probably just get passed to taskset.

I'm still not sure about the details here. `taskset` is a specific (Linux-only) CLI utility for running a command with a certain CPU affinity set. It might not even be usable in all situations (I don't fully understand the permission issues) Some of our tools, such as the Python ones, don't run any subcommands, so they cannot use `taskset`. There are related (Linux-only) syscalls for setting thread affinity that are to some extent equivalent to what `taskset` does. Do you envision tools using those?

> Some tools may also choose to accept a boolean --multithread or --multithreaded argument

I would prefer a single option, where `--multithreaded` switches on multithreaded execution using as many cores as the tool thinks is a good idea, and `--multithreaded=N` would do the same but use only up to N cores.

Generally, the whole business of limiting the number of cores is a technical minefield. I'm not at all confident in my ability to implement it correctly, or to understand the portability issues (on non-Linux) or permission issues (on HPC systems) that may occur. It *may* all work without much trouble, but it is outside my experience to say *whether* it will (except it will for sure not work on macOS).

My preference would be to *only* have `--multithreaded`, as currently implemented in this PR. If a user wishes to limit the number of cores used for multithreaded execution, then they can manually use `taskset` before running `gradbench` (or make it part of the `--tool` command), *or* use whatever mechanism is appropriate for their execution environment. E.g. on an HPC system, you would tell Slurm how many cores you want to have allocated. I think this will serve most users. Only people who are specifically interested in studying parallel scalability of AD tools need to vary the thread count. Most others just need to run a tool as fast as possible.

samestep:
> `taskset` is a specific (Linux-only) CLI utility for running a command with a certain CPU affinity set.

Ah I see, you're saying Linux-only is an issue for running outside Docker on macOS? I hadn't thought of that but it does make sense. Given that we're probably going to have "all cores allowed" be the default, seems probably fine for the `--cores` arg only work on Linux, so on macOS you'd need to do it inside Docker? Or what are your thoughts?

> It might not even be usable in all situations (I don't fully understand the security model).

Sure, if there are more details I'm not aware of that make it infeasible, then that's another matter.

> Some of our tools, such as the Python ones, don't run any subcommands, so they cannot use `taskset`. There are related (Linux-only) syscalls for setting thread affinity that are to some extent equivalent to what `taskset` does. In some languages (such as Python) these are exposed in the standard library. Do you envision tools using those?

For Python, I was envisioning doing the same thing as for C++ and others: have a pretty generic wrapper script that processes args like `--cores`, and then have that wrapper script call the actual Python tool script as a subcommand. No need to do everything all in the same process. And actually, then, we wouldn't even really have the "Linux-only" issue for running outside Docker: we can tell people to just run the actual Python tool script directly if they don't care about setting `--cores`, so then it'd all be in one process.

> I would prefer a single option, where `--multithreaded` switches on multithreaded execution using as many cores as the tool thinks is a good idea, and `--multithreaded=N` would do the same but use only up to N cores.

This I'm very strongly against, because it violates standard conventions for CLI args: specifically, if a CLI flag takes an argument, it is pretty much always allowed to do it via either a space or an equals sign, This means that if `--multithreaded` can take _either_ zero or one argument, then you'd need a nonstandard arg parser, because if you do something like `--multithreaded --some-other-arg`, then it wouldn't know whether or not that `--some-other-arg` is its input or just a separate argument. Sure you could do that by first checking if it's an integer or not, but again, that's very nonstandard. Let's not do this.

> Generally, the whole business of limiting the number of cores is a technical minefield. I'm not at all confident in my ability to implement it correctly, or to understand the portability issues (on non-Linux) or permission issues (on HPC systems) that may occur. It _may_ all work without much trouble, but it is outside my experience to say _whether_ it will (except it will for sure not work on macOS).

Right, that's fair. But again, if we go with the wrapper script approach, that seems fine, right? As long as we're pretty sure it'll work inside Docker, and we have a way to bypass it outside Docker, I don't really see the downside.

> My preference would be to _only_ have `--multithreaded`, as currently implemented in this PR. If a user wishes to limit the number of cores used for multithreaded execution, then they can manually use `taskset` before running `gradbench` (or make it part of the `--tool` command), _or_ use whatever mechanism is appropriate for their execution environment. E.g. on an HPC system, you would tell Slurm how many cores you want to have allocated. I think this will serve most users. Only people who are specifically interested in studying parallel scalability of AD tools need to vary the thread count. Most others just need to run a tool as fast as possible.

Sure, I'm fine with doing only `--multithreaded` in this PR, since it seems like we both agree that semantically we'd want it to be separate from `--cores` anyway. We can leave `--cores` as a followup if we decide that it's worth it to do later.

samestep:
Should I mark this PR as resolving #303?

athas:
Yes, I suppose it does, although making all tools multi threaded where possible is going to be a long path - and this PR won't do it all.

samestep:
Sounds good; do you think that we should link these together so that gets closed when we merge this, and then open more specific followup issues for individual tools that should be parallelized?

athas:
I think I have done so now.

athas:
I added some documentation. My plan is to fix (and optimise) `llsq::primal``, then mark this as ready for review.

athas:
I think this is ready now. I attempted to implement my modifications such that performance of the sequential case is unaffected, but we do not have good tooling for actually verifying it. I am *pretty sure* it's fine, as most changes are just the addition of OpenMP pragmas, which are ignored in the sequential case.

README.md
```diff
@@ -234,6 +235,33 @@ PYTHONPATH=python/gradbench/:$PYTHONPATH ./gradbench repo run --eval "$ python3
...
+### Multithreading
+
+By default, tools use only a single thread. You can ask them to use multiple
+(CPU) threads, if possible, by passing the option `--multithread`. If you use
+`gradbench repo tool`, then remember to pass `--` before any tool-specific
+options. Example:
+
+```sh
+./gradbench repo tool manual -- --multithread
+```
```
samestep, line 247: I think using `repo tool` here is out of place now that #522 has edited all this documentation to use `repo run` instead. Can you update this section to use `repo run` as well?

README.md
```diff
@@ -234,6 +235,33 @@ PYTHONPATH=python/gradbench/:$PYTHONPATH ./gradbench repo run --eval "$ python3
...
+use only a single thread. Some tools may be able to multithread their primal
+code, but not their differentiated code. Some tools may fail to work entirely.
+An eval documents to which extent it can be parallelised in its _Commentary_,
+and a tool similarly documents which of its implementations have been
+parallelised in its _Commentary_.
+
+The following tools have at least partial support for multithreaded execution of
+their differentiated functions:
+
+- [finite](tools/enzyme)
```
samestep, line 259:
```suggestion
- [enzyme](tools/enzyme)
```

tools/enzyme/README.md
```diff
@@ -5,3 +5,23 @@
 Enzyme needs to be compiled against a specific version of LLVM. You are strongly advised to use the Dockerfile for this one.
 
 [Enzyme]: https://enzyme.mit.edu/
+
+## Commentary
+
+### Multithreading
+
+The approach here is very similar to the one for [manual](../manual),
```
samestep, line 13: This sort of relative link is error-prone; please replace it with one using a leading slash.

tools/enzyme/README.md
```diff
@@ -5,3 +5,23 @@
...
+except that the derivative functions are (obviously) produced with
+Enzyme. In particular, the primal functions use the [C++ reference
+implementations][../../cpp/gradbench/evals] - see the Commentary for
```
samestep, line 16: This link is broken; please fix it and also replace the relative link with one using a leading slash.

tools/manual/README.md
```diff
@@ -29,3 +29,25 @@ implementation of a tool does not do. It is not a goal that the
 It is not expected that we will be able to implement
 hand-differentiated versions of all evals. AD is after all most useful
 for those cases where hand-differentiation is impractical.
+
+### Multithreading
+
+The primal functions use the [C++ reference
+implementations][../../cpp/gradbench/evals] - see the Commentary for
```
samestep, line 36: Similar to above, this link is broken and should also be replaced with one using a leading slash.

samestep (changes requested):
Thanks for doing this!

The main clarification I'd like to see is about the interface. Specifically:

1. What exactly is the name of this CLI argument supposed to be? I see `--multithread`, `--multithreaded`, and `--multithreading` written in various different places. _Technically_ Python's `argparse` accepts the first when the second is the actual name, but then the third is still just wrong. Please pick one of these three and go through and consistently use it both in the implementations and in the documentation.

2. I see that this PR makes no edits to our CI setup. Does this mean that what is run in CI will basically remain the same after this is merged? In the original issue #303 we talked about having CI run with multiple configurations, e.g. both single-core and multi-core. Do we still want to do that? If not, which one should be the configuration we run in CI for the time being? Do we want to open a separate GitHub issue about using multi-threading in CI specifically?

(As a minor aside: looks like lines in this PR are still being wrapped at 70 columns? Is that intentional? I seem to remember discussing switching to wrapping at 80 columns, but I can't remember the conclusion of that conversation.)

(Also, having to manually flag all these broken links is really making me think something like #333 might be a good idea...)

athas:
1. I waffled about and decided on `--multithreaded`, but I don't have a good argument for why (which is why I could apparently never remember it when working on different places).
2. Yes. For this PR I wanted to focus purely on the implementation, because it is a simpler question, and I am somewhat lazy. Following this philosophy, I also suggest that we keep CI the way it currently is (sequential execution) and establish a subcommittee for deciding how (or whether) to run multithreaded benchmarks in CI. I think that does not become very interesting until more tools have been multithreaded, however.

samestep:
> 2. Yes. For this PR I wanted to focus purely on the implementation, because it is a simpler question, and I am somewhat lazy. Following this philosophy, I also suggest that we keep CI the way it currently is (sequential execution) and establish a subcommittee for deciding how (or whether) to run multithreaded benchmarks in CI. I think that does not become very interesting until more tools have been multithreaded, however.

OK, makes sense!

In that case, why do the `llsq` and `lse` rows of the CI grid look so different [before](https://gradben.ch/?date=2025-04-29) this PR and [after](https://gradben.ch/?commit=3e976d189077185fb73e3f89e6bd2853adcba3b4) it? If multithreading isn't being used in CI yet for this PR, I'd expect those results to look the same.

athas:
`lse` looks like TensorFlow jitter, but it seems I made the C++ primal of `llsq` a lot slower by accident. I will investigate.

athas:
Oh right, it predates the optimisation I did in #538. I will also do that to this version.

athas:
Should be fixed now.

samestep (approved):
Looks good to me!

Oddly, when I ran the `repo run` example this PR adds to the README, it didn't seem noticeably faster on my MacBook.

athas:
Not even for `jacobian`? I don't see `objective` benefiting much (maybe it would on larger workloads), but I get a nontrivial speedup on `jacobian`.

samestep:
Oh never mind, sorry; I was looking at the numbers backward. With `--multithreaded`, `jacobian` is indeed significantly slower on the smaller examples, but it's about 40% faster for `64_100_1000`; is that more in line with what you see?

athas:
I see about a 4.5x speedup on that workload, but that is on a 16 core CPU.

PR #478, opened 2025-04-08
New eval: logsumexp

athas:
This is not a very interesting eval - in fact, it already exists as a subfunction in `gmm`. However, it is a distinct benchmark in the [Enzyme benchmark suite](https://github.com/EnzymeAD/Enzyme/blob/main/enzyme/benchmarks/ReverseMode/logsumexp/logsumexp.cpp), and it is both realistic (occurs in real AD use cases) and pretty easy to implement. I might actually be able to implement it for all of our tools.


athas:
I think I have now implemented this with everything I am able to. Missing:

* `floretta`: I can produce the WASM code, but I don't know enough about Floretta itself to call it.
* `scilean`: My keyboard doesn't have those keys. Leaving this one for @lecopivo.
* `tapenade`: our tooling for this is pretty dire - it's specialised for `hello`, and I don't have the fortitude to rewrite it right now.

samestep (approved):
Thanks, yeah this seems good! I'll implement for Floretta in a followup PR.

samestep:
No need to shorten the eval name unless you really want to, I would say.

athas:
You're OK with the name? It doesn't look great in the console output:

```
  [0] start logsumexp (enzyme-jl)
  [1] def   logsumexp                              17ms ✓
  [2] eval  logsumexp::p... n=10000               549ms ~       121ms evaluate ✓
  [4] eval  logsumexp::g... n=10000             4.146 s ~     4.052 s evaluate ✓
  [6] eval  logsumexp::p... n=100000               65ms ~         3ms evaluate ✓
  [8] eval  logsumexp::g... n=100000              111ms ~         8ms evaluate ✓
 [10] eval  logsumexp::p... n=1000000             598ms ~        30ms evaluate ✓
 [12] eval  logsumexp::g... n=1000000           1.353 s ~        64ms evaluate ✓
```

samestep:
Hmm, should we just make that part of the output a bit wider maybe?

athas:
That would be a solution, but it's nice that the output is usually not too wide. We could also suppress the module name - usually it's obvious from the context.

samestep:
I'm totally fine with just shortening the name like you suggested; these other alternatives maybe aren't good enough to justify.

athas:
I once realised that [I have never regretted making a name too short](https://futhark-lang.org/blog/2021-01-11-no-regrets.html#short-names) but I have often regretted making it too long. I will shorten this to `lse`.

Issue #475, opened 2025-04-07
Add more Julia AD systems with DifferentiationInterface?


gdalle:
Hi, lead dev of DifferentiationInterface speaking. Just saying I'd be happy to lend a hand in using DI as a common entry point to a couple Julia AD systems.
About mutation support, DifferentiationInterface has an internal mechanism to decide whether the backend supports it or not, it could be made public and used here if relevant:

```julia
julia> using DifferentiationInterface, Enzyme, Zygote

julia> DifferentiationInterface.inplace_support(AutoEnzyme())
DifferentiationInterface.InPlaceSupported()

julia> DifferentiationInterface.inplace_support(AutoZygote())
DifferentiationInterface.InPlaceNotSupported()
```

Related:
- #464 
- #463 

Ping @vchuravy @athas

gdalle:
In particular I think @willtebbutt will want to try Mooncake

athas:
Addressed in #575, although there are still many benchmarks missing.

PR #479, opened 2025-04-09
Support more Julia AD backends through DifferentiationInterface

gdalle:
This PR adds the necessary infrastructure to support every Julia AD system at once, using [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl) (DI):

- Add utilities to `julia/GradBench` to pick a backend
- Adjust the `ode` and `hello` examples to use that backend with DI
- Add / modify tool implementations for the most widely used libraries in the Julia ecosystem:
  - [ ] Leave [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) alone (it should _not_ be switched to DI[^1])
  - [x] Add [ForwardDiff.jl](https://github.com/JuliaDiff/ForwardDiff.jl)
  - [x] Add [Mooncake.jl](https://github.com/compintell/Mooncake.jl)
  - [x] Add [ReverseDiff.jl](https://github.com/JuliaDiff/ReverseDiff.jl)
  - [x] Modify [Zygote.jl](https://github.com/FluxML/Zygote.jl)

Ping: @athas @vchuravy @willtebbutt

Related:

- #475

Can I test this locally without using `./gradbench`? When I run e.g. `tools/zygote/run.jl` (as suggested in #476), it hangs and expects an input in the stdin but I don't know what that input should be.

[^1]: Despite my best efforts, some Enzyme devs are worried that going through DI might misrepresent the performance or correctness of their library

### DIFF
--- a/julia/GradBench/src/GradBench.jl
+++ b/julia/GradBench/src/GradBench.jl
@@ -13,9 +13,9 @@ function register!(mod, functions)
 end
 
 # Avoid measuring dispatch overhead
-function measure(func::F, arg) where {F}
+function measure(func::F, args::Vararg{Any,N}) where {F,N}
     start = time_ns()
-    ret = func(arg)
+    ret = func(args...)
     done = time_ns()
     return ret, done - start
 end
@@ -23,6 +23,7 @@ end
 function run(params)
     mod = DISPATCH_TABLE[params["module"]]
     func = mod[params["function"]]
+    backend = get(mod, "backend", nothing)
     arg = params["input"]
     min_runs = get(params, "min_runs", 1)
     min_seconds = get(params, "min_seconds", 0)
@@ -37,7 +38,11 @@ function run(params)
     i = 1
     ret = nothing
     while i <= min_runs || elapsed_seconds <= min_seconds
-        ret, t = measure(func, arg)
+        if isnothing(backend)
+            ret, t = measure(func, arg)
+        else
+            ret, t = measure(func, arg, backend)
+        end
         push!(timings, Dict("name" => "evaluate", "nanoseconds" => t))
         elapsed_seconds += t / 1e9
         i += 1
--- a/julia/GradBench/src/GradBench.jl
+++ b/julia/GradBench/src/GradBench.jl
@@ -23,7 +23,11 @@ end
 function run(params)
     mod = DISPATCH_TABLE[params["module"]]
     func = mod[params["function"]]
-    backend = get(mod, "backend", nothing)
+    if func isa Tuple
+        func, backend = func
+    else
+        backend = nothing
+    end
     arg = params["input"]
     min_runs = get(params, "min_runs", 1)
     min_seconds = get(params, "min_seconds", 0)
@@ -41,7 +45,7 @@ function run(params)
         if isnothing(backend)
             ret, t = measure(func, arg)
         else
-            ret, t = measure(func, arg, backend)
+            ret, t = measure(func, backend, arg)
         end
         push!(timings, Dict("name" => "evaluate", "nanoseconds" => t))
         elapsed_seconds += t / 1e9
--- a/julia/GradBench/src/benchmarks/ode.jl
+++ b/julia/GradBench/src/benchmarks/ode.jl
@@ -23,7 +23,7 @@ function ode_fun(x, y, z)
     end
 end
 
-function primal!(y::Vector{T}, x::Vector{T}, s::Real) where {T}
+function primal!(y::AbstractVector{T}, x::AbstractVector{T}, s::Real) where {T}
     tf = T(2)
     h = tf / T(s)
 
@@ -60,13 +60,13 @@ function primal!(y::Vector{T}, x::Vector{T}, s::Real) where {T}
     return nothing
 end
 
-function primal(x::Vector, s::Real)
+function primal(x::AbstractVector, s::Real)
     y = similar(x)
     primal!(y, x, s)
     return y
 end
 
-function gradientlast(backend::AbstractADType, x::Vector, s::Real)
+function gradientlast(backend::AbstractADType, x::AbstractVector, s::Real)
     return DI.gradient(last ∘ primal, backend, x, DI.Constant(s))
 end
 
--- a/julia/GradBench/src/benchmarks/ode.jl
+++ b/julia/GradBench/src/benchmarks/ode.jl
@@ -65,7 +65,7 @@ function primal(x::Vector, s::Real)
     return y
 end
 
-function gradientlast(x::Vector, s::Real, backend::AbstractADType)
+function gradientlast(backend::AbstractADType, x::Vector, s::Real)
     return DI.gradient(last ∘ primal, backend, x, DI.Constant(s))
 end
 
@@ -80,9 +80,9 @@ function primal_from_message(message::Dict)
     return primal(x, s)
 end
 
-function gradientlast_from_message(message::Dict, backend::AbstractADType)
+function gradientlast_from_message(backend::AbstractADType, message::Dict)
     x, s = parse_input(message)
-    return gradientlast(x, s, backend)
+    return gradientlast(backend, x, s)
 end
 
 end # module ode
--- a/julia/GradBench/src/benchmarks/ode.jl
+++ b/julia/GradBench/src/benchmarks/ode.jl
@@ -12,10 +12,8 @@
 
 module ODE
 
-struct Input
-    x::Vector{Float64}
-    s::Int
-end
+using ADTypes: AbstractADType
+import DifferentiationInterface as DI
 
 function ode_fun(n, x, y, z)
     z[1] = x[1]
@@ -24,40 +22,67 @@ function ode_fun(n, x, y, z)
     end
 end
 
-function primal(n, xi::Vector{T}, s, yf::Vector{T}) where {T}
+function primal!(y::Vector{T}, x::Vector{T}, s::Real) where {T}
     tf = T(2)
     h = tf / T(s)
 
-    k1 = Vector{T}(undef, n)
-    k2 = Vector{T}(undef, n)
-    k3 = Vector{T}(undef, n)
-    k4 = Vector{T}(undef, n)
-    y_tmp = Vector{T}(undef, n)
+    k1 = similar(y)
+    k2 = similar(y)
+    k3 = similar(y)
+    k4 = similar(y)
+    y_tmp = similar(y)
 
-    yf .= T(0)
+    y .= T(0)
 
     for _ in 1:s
-        ode_fun(n, xi, yf, k1)
+        ode_fun(n, x, y, k1)
 
-        for i in 1:n
-            y_tmp[i] = yf[i] + h * k1[i] / T(2)
+        for i in eachindex(y_tmp, y, k1)
+            y_tmp[i] = y[i] + h * k1[i] / T(2)
         end
-        ode_fun(n, xi, y_tmp, k2)
+        ode_fun(n, x, y_tmp, k2)
 
-        for i in 1:n
-            y_tmp[i] = yf[i] + h * k2[i] / T(2)
+        for i in eachindex(y_tmp, y, k2)
+            y_tmp[i] = y[i] + h * k2[i] / T(2)
         end
-        ode_fun(n, xi, y_tmp, k3)
+        ode_fun(n, x, y_tmp, k3)
 
-        for i in 1:n
-            y_tmp[i] = yf[i] + h * k3[i]
+        for i in eachindex(y_tmp, y, k3)
+            y_tmp[i] = y[i] + h * k3[i]
         end
-        ode_fun(n, xi, y_tmp, k4)
+        ode_fun(n, x, y_tmp, k4)
 
-        for i in 1:n
-            yf[i] += h * (k1[i] + T(2) * k2[i] + T(2) * k3[i] + k4[i]) / T(6)
+        for i in eachindex(y, k1, k2, k3, k4)
+            y[i] += h * (k1[i] + T(2) * k2[i] + T(2) * k3[i] + k4[i]) / T(6)
         end
     end
+    return nothing
+end
+
+function primal(x::Vector, s::Real)
+    y = similar(x)
+    primal!(y, x, s)
+    return y
+end
+
+function gradientlast(x::Vector, s::Real, backend::AbstractADType)
+    return DI.gradient(last ∘ primal, backend, x, DI.Constant(s))
+end
+
+function parse_input(message::Dict)
+    x = convert(Vector{Float64}, message["x"])
+    s = message["s"]
+    return x, s
+end
+
+function primal_from_message(message::Dict)
+    x, s = parse_input(message)
+    return primal(x, s)
+end
+
+function gradientlast_from_message(message::Dict, backend::AbstractADType)
+    x, s = parse_input(message)
+    return gradientlast(x, s, backend)
 end
 
 end # module ode
--- a/julia/GradBench/src/benchmarks/ode.jl
+++ b/julia/GradBench/src/benchmarks/ode.jl
@@ -15,7 +15,8 @@ module ODE
 using ADTypes: AbstractADType
 import DifferentiationInterface as DI
 
-function ode_fun(n, x, y, z)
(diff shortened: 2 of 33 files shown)


athas:
> Can I test this locally without using ./gradbench? When I run e.g. tools/zygote/run.jl (as suggested in https://github.com/gradbench/gradbench/pull/476), it hangs and expects an input in the stdin but I don't know what that input should be.

"Running directly" still means having to go through the `gradbench` CLI - it just means you don't have to run it inside Docker. The problem is that the workloads are usually generated dynamically by the evals, so there is no "input file" one could pass to the program.

gdalle:
Should I get rid of ForwardDiff? It is a forward-mode backend, and I'm unclear whether all of your benchmarks are reverse-mode-oriented or not

athas:
Not all of them, but certainly the majority. GradBench works fine with extremely slow tools, so I would keep it.

gdalle:
Sorry there was an issue with Julia dependency management, that's why CI failed. Should be fixed by the latest commit, can you reapprove the workflow @athas ?

gdalle:
> GradBench works fine with extremely slow tools, so I would keep it.

So apparently ForwardDiff and ReverseDiff timed out after 10 minutes on the largest ODE instance, and this made CI fail. Is there anything I should code differently for it to succeed?

On the other hand, I'm not sure I understand the [error for Mooncake](https://github.com/gradbench/gradbench/actions/runs/14353623717/job/40240473692?pr=479):

```
invalid JSON from tool: EOF while parsing a value at line 1 column 0
```

athas:
> So apparently ForwardDiff and ReverseDiff timed out after 10 minutes on the largest ODE instance, and this made CI fail. Is there anything I should code differently for it to succeed?

Change the line in `evals.txt` to read

    ode timeout

> On the other hand, I'm not sure I understand the [error for Mooncake](https://github.com/gradbench/gradbench/actions/runs/14353623717/job/40240473692?pr=479):

This usually means the tool crashed and produced empty output.

gdalle:
Is there any way to see where it crashed and the error message (e.g. out of memory or sth)?

athas:
The GI output *should* contain the stderr of the tool. However, if the problem is OOM, then my experience is that it just gets killed, and the tooling doesn't properly notice or log this.

samestep:
Right, all stderr gets sent transparently to the terminal, so if the tool is OOMing with nothing printed to stderr, seems like there's nothing the intermediary could do about that? Although I suppose we could modify the generic error message to say that this means there was an empty output and speculate on what the cause might be.

vchuravy:
> Add utilities to julia/GradBench to pick a backend

My idea behind GradBench was that it would just contain the primal codes + measurement and run harness. 

Every AD framewrok should then implement an appropriate gradient method.
I think I would prefer an approach where we have a `tools/differentiationinterface-jl` and we have a way of specifying (perhaps in evals.txt) the backend that should be used for `DI`.

So you could say:

```
ode AutoEnzyme
ode AutoReverseDiff
```

This doesn't exist yet, but maybe isn't too hard?

gdalle:
> I think I would prefer an approach where we have a tools/differentiationinterface-jl and we have a way of specifying (perhaps in evals.txt) the backend that should be used for DI.

From what I understand, this is at odds with the current structure of `gradbench` as a repo, because DI is not a tool: it is an entry point for half a dozen different tools.

I agree that there is code duplication between the individual tools called by DI, but as things currently stand I'm not sure we can avoid it. And if we can't, then the (very lightweight) DI bindings should probably live in `julia/GradBench`?

vchuravy:
>  because DI is not a tool: it is an entry point for half a dozen different tools.

I would say it is a tool and it is an entry point for many different backends. As an example DI could feasibly implement a "better" Jacobian function, and I for one would be interested in comparing the performance of DI as a tool vs directly using Enzyme/Zygote.

So I don't see my proposal at odds with the structure of GradBench.

I think generally there is an open question of how to write ["variants" of a benchmark](https://github.com/gradbench/gradbench/pull/464#issuecomment-2781505235) 

> Not to mention that this isn't really enough, because backend objects from ADTypes can have additional options which would be awkward to encode as strings.

You can have an arbitrary mapping of "(problem, options)" to a problem specification,
`Dict(("ode", "enzyme") => AutoEnzyme())`.

> Despite my best efforts, some Enzyme devs are worried that going through DI might misrepresent the performance or correctness of their library 

I don't generally worry about DI missrepresenting Enzyme correctness, but I would argue that the goal of GradBench is to show the idiomatic application of an AD tool onto a problem. Characterizing the performance with and without DI is at least for me of import, as is contrasting the benefits DI unified user interface brings with how the backends express problems.

So, if options are not possible, I would indeed argue for something like:

```
tools/enzyme-jl
tools/zygote-jl
tools/differentiationinterface-enzyme-jl
tools/differentiationinterface-zygote-jl
```

or 

```
tools/enzyme-jl
tools/zygote-jl
tools/differentiationinterface/enzyme-jl
tools/differentiationinterface/zygote-jl
```

samestep:
I'll split my perspective into two parts. The first part is more normative, whereas the second part is merely a suggestion.

1. **_As a GradBench maintainer,_** I'm currently not really interested in complicating our simple evals/tools model by letting an eval appear multiple times in a tool's `evals.txt` file with different parameters.

   - It _is_ totally fine to give a tool command-line parameters that you can pass to run it in different configurations that aren't the default used by CI. Currently `CONTRIBUTING.md` only documents this for evals (which actually answers @gdalle's question "Is there a way to only run a given tool up to a certain size and no further?"), but it's also allowed and encouraged for tools as well.

   - It's _also_ fine to split two variants/configurations of a tool into multiple "tools" under the `tools/` directory, if we consider them important enough to run them both in CI; for instance, we already have both `tools/enzyme` and `tools/enzyme-jl`, even though in some sense they're both "the same", because in another important sense they're different. This is a judgment call.

   - As I've mentioned before, having two different folders under `tools/` doesn't actually require you to duplicate code. For instance, all the Python tools share a _single_ codebase under `python/`. It seems reasonable to do the same for Julia, so then there should be no need to actually _duplicate_ any code, but of course each tool should feel free to make its own different versions of things if it's more idiomatic for that tool.

2. **_As a Julia outsider,_** I agree that it'd be interesting to see the difference in using DifferentiationInterface vs not. As I wrote above, though, that is not going to be represented on the "primary" GradBench eval/tool matrix by DifferentiationInterface itself being a single "tool" or column, because the whole point of DifferentiationInterface is to support many different backends. Because no single backend makes sense even as the "default", I do not think that there should be a `tools/differentiationinterface` at all.

   - Instead, just like every other tool in this repo, each Julia tool must choose whether or not using DifferentiationInterface is the default configuration for being run in CI, but is totally free to also provide other configurations as well through CLI flags. For consistency, I'd propose that for Julia tools in this repo that support being run both with and without DifferentiationInterface, there be CLI flags `--differentiation-interface` and `--no-differentiation-interface` to let people choose which configuration is run.

   - Obviously this means that not all possible configurations of all possible tools are going to be run in CI, but that's already the case outside of Julia; not every interesting question is going to be answerable by the set of things that is run in GradBench CI. However, we do want to enable people to answer other sorts of questions by running GradBench locally with different configurations of tools; if there are currently missing features that would be important for doing that, please feel free to open a GitHub issue!

athas:
Unfortunately any configuration options passed to a tool will not be visible in the log file. I think we should specify some kind of convention (perhaps a field in the response to the `start` message) where a tool can indicate any nonstandard configuration. This would help keep log files understandable on their own.

samestep:
@athas the `StartResponse` type in the protocol already has a `config` field which you added in #231; does this not already serve that purpose?

https://github.com/gradbench/gradbench/blob/b0ef028005b23685bc0061727820ffba67c9c77b/CONTRIBUTING.md?plain=1#L348-L351

athas:
Oh right, I guess I forgot that we already took care of it. We should make sure that tools make use of it when applicable. I think `futhark` is the only tool with configuration options, and it does use it.

athas:
For tools that expect to fail on some eval, you can put

```
ode failure
```

in `evals.txt`. Of course, it's best to make them work. But when I run `mooncake-jl` on `ode`, it certainly does eat all of my memory.

athas:
Note that even if a tool fails (or times out), the results still get plotted. It's an explicitly expected and acceptable method for tools to handle some evals.

gdalle:
While the `gradbench` maintainers hash out non-standard configurations for tools, I think we need to decide what we pick for the standard.
My inclination is to keep DI as the default for most Julia backends, and leave Enzyme as the exception (which this PR does) due to its more advanced performance optimizations.
Sure, we may leave some performance on the table (presumably not much), but it will allow quicker implementation and comparison across backends for the time being.

athas:
I agree. I also find it likely that DI-as-default will make it easier to have a large number of implementations. People with a particular interest in the underlying libraries can then implement tuned versions as they deem it necessary.

vchuravy:
It is up to the Grad bench maintainers of course, but my understanding was that GradBench was aiming for idiomatic examples and implementations.

For me DI is not the idiomatic frontend to tools like Zygote, ForwardDiff, or Enzyme. It might be the idiomatic frontend for Mooncake?

DI is a very powerful and useful tool, especially when experimenting with different AD frameworks, but it also an additional layer on top of the AD frameworks. 

I still strongly feel that there is a separation of concerns here. As an example taking the Zygote change, DI is now hiding/obfuscating the details of how to use Zygote which might be of interest to a student of AD frameworks.

So we should absolutely have DI variants and examples, but we should also the idiomatic variants for each tool.

samestep:
Personally I don't see any disagreement here? @gdalle is saying we should start by using DI to save time a beginning for supporting a bunch of tools at once, which I agree with; and @vchuravy is saying that some tools shouldn't use DI by default in GradBench because it's not their most idiomatic usage scenario, which I also agree with. Or am I missing a key point of friction?

julia/GradBench/src/benchmarks/ode.jl
```diff
@@ -12,52 +12,78 @@
...
+end
+
+function primal(x::AbstractVector, s::Real)
+    y = similar(x)
+    primal!(y, x, s)
+    return y
+end
+
+function gradientlast(backend::AbstractADType, x::AbstractVector, s::Real)
+    return DI.gradient(last ∘ primal, backend, x, DI.Constant(s))
```
vchuravy, line 70: This is a different complexity than the C++ implementation.

The C++ benchmark essentially is doing a vJp with an vector, for Enzyme the variant here would be faster ( if activity analysis is successful)

gdalle: I am aware, I went for the easiest implementation first. DI can also support a `pullback` on the last primal component, but for ForwardDiff and ReverseDiff it will be much slower, which is why I didn't consider it at first.

vchuravy:
I think my objections are:

1. I would prefer leaving Zygote as is an not using DI as interface for it.
2. A structural debate whether or not the GradBench.jl package should contain the DI gradient implementation or if they should life somewhere else.

Regarding 2, my long term aim was to make GradBench.jl usable by Enzyme to provide a source of benchmark and I intentionally did not add any dependency to GradBench.jl for an AD tool, and instead wrote the gradient definition explicitly out in the run_*.JL file.

I prefer this explicit style since it allows the reader to contrast different AD frameworks.

vchuravy:
Also I don't feel the cost is in adding more Julia based AD tools the  writing the gradient wrappers is often straightforward, the cost right now is more in porting the benchmarks and deciding the style of the benchmark so that it is compatible with your AD framework of choice

gdalle:
> Regarding 2, my long term aim was to make GradBench.jl usable by Enzyme to provide a source of benchmark and I intentionally did not add any dependency to GradBench.jl for an AD tool, and instead wrote the gradient definition explicitly out in the run_*.JL file.

Note that my addition of the DI machinery does not prevent you from doing that. DI is as light a dependency as they come (it only requires ADTypes and LinearAlgebra), so if it's loading time you're worried about, anything you do with Enzyme will absolutely dwarf the overhead of additional DI code you don't use.

gdalle:
> Also I don't feel the cost is in adding more Julia based AD tools the writing the gradient wrappers is often straightforward, the cost right now is more in porting the benchmarks and deciding the style of the benchmark so that it is compatible with your AD framework of choice

It's great that you don't feel this cost, but someone has to feel it while writing and maintaining these bindings. That's the whole idea behind DI. There are a dozen different syntaxes to call gradients in Julia. Sure, individually they don't cost a lot to understand and use, but it still requires checking out every single documentation. And perhaps we can't count on a dedicated Julia user keeping all of this up to speed for every change in the backends, whereas it will by definition be kept up to speed in DI cause everyone depends on it.

gdalle:
Would it feel cleaner to you @vchuravy if I were to make a `julia/GradBenchDI` which depends on `GradBench` and implements the additional DI machinery?

julia/GradBench/src/benchmarks/ode.jl
vchuravy: Can you add a comment?

vchuravy:
> It really depends what you define by idiomatic.

Does the Readme of the package recommend it?

> Would it feel cleaner to you @vchuravy if I were to make a julia/GradBenchDI which depends on julia/GradBench and implements the additional DI machinery?

Yes, but that also feels over-engineered. Lets keep it as is. We can rework things later if it becomes a problem. I think I was just looking at the main API and got annoyed to see the DI backend there. Maybe we can just use closures?

tools/forwarddiff-jl/run_hello.jl
```diff
@@ -0,0 +1,14 @@
+module Hello
+
+using ADTypes: AutoForwardDiff
+import ForwardDiff
+import GradBench
+
+GradBench.register!(
+    "hello", Dict(
+        "square" => GradBench.Hello.square,
+        "double" => (GradBench.Hello.double, AutoForwardDiff()),
```
vchuravy, line 10: Can we instead just use a closure here?

julia/GradBench/src/GradBench.jl
```diff
@@ -13,16 +13,21 @@ function register!(mod, functions)
...
-    ret = func(arg)
+    ret = func(args...)
     done = time_ns()
     return ret, done - start
 end
 
 function run(params)
     mod = DISPATCH_TABLE[params["module"]]
     func = mod[params["function"]]
+    if func isa Tuple
```
vchuravy, line 26: This code can be removed now, right?

athas:
I wrote a script for extracting the GradBench input data: https://github.com/gradbench/gradbench/blob/scripts/scripts/gradburglar

Maybe this can help with testing and debugging.

julia/GradBench/src/GradBench.jl
gdalle: No because it enables the DI versions?

tools/forwarddiff-jl/run_hello.jl
gdalle: Not if you want to avoid the performance pitfalls of closures, e.g. with DI + Enzyme (which at the moment isn't measured but it would be unfair to set it up with a disadvantage)

julia/GradBench/src/benchmarks/ode.jl
```diff
@@ -12,52 +12,78 @@
...
+end
+
+function parse_input(message::Dict)
```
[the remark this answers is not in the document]
athas: Regarding this, the Julia benchmark modules I have recently added define a type `Input` and a function `input_from_json` that produces an `Input` from a value representing the JSON input. The `input_from_json` function is currently called directly by the benchmark functions, but it is exactly the thing that should be lifted out and only done once. (Feel free to come up with a more Julia-appropriate naming scheme.) A similar thing should be done for the output - sometimes it has to be formatted in benchmark-specific ways, which we would like not to measure.

tools/forwarddiff-jl/run_hello.jl
vchuravy: I don't quite understand? My propsal is essentially to define:

```
function double(x)
      GradBench.Hello.double(AutoForwardDiff(), x)
end
```

But with an anonymous function?

gdalle:
Closing this to reopen a new one now that the underlying structure has changed

PR #482, opened 2025-04-09
New tool: ad.hpp

athas:


athas:
This is missing `ht` (because I don't know how to do forward mode) and `kmeans` (because I don't know how to do higher order). Those can always be implemented later, so I think it's good enough to go.

samestep:
Ah sorry I forgot to review this. Looks great though, thanks for adding it!

Issue #491, opened 2025-04-15
Inverse of sparse Hessian is NOT the elementwise reciprocal.


aj-fleming:
https://github.com/gradbench/gradbench/blob/1c2a0364010e8a7c9e93aa2b8c667ba50689134e/evals/kmeans/README.md?plain=1#L21

No it isn't, unless the Hessian is sparse _diagonal_.

athas:
Fixed by #492.

PR #492, opened 2025-04-15
Hessian is sparse diagonal.

aj-fleming:
Corrected README in kmeans eval.


aj-fleming:
...fixes my complaint in #491

PR #493, opened 2025-04-15
Implement gmm with zygote

athas:
This PR adds an implementation of `gmm` using `zygote`, which is largely ported from ADBench. It makes use of the design introduced in #464 by @vchuravy, in the sense that there is a general `GMM` module in the `GradBench` package that is not Zygote-specific. My hope is that this will be usable by the work done by @gdalle in #479.

The implementation has the same problem as the other Julia benchmarks, in that the input unpacking is measured, and the runtime of the first workload is inflated due to JIT compilation.

Performance is not good, but it [looks like that was also the case in ADBench](https://github.com/microsoft/ADBench/pull/98#issuecomment-535943890).


tools/zygote/hello.jl
```diff
@@ -0,0 +1,14 @@
+module hello
```
vchuravy, line 1: This seems to be accidentially checked-in.

tools/zygote/run_gmm.jl
```diff
@@ -0,0 +1,91 @@
...
+        for i in 1:size(Qs, 3)
+            for j in 1:size(Qs, 1)
+                Δ′[j, j, i] *= exp(Qs[j, j, i])
+            end
+        end
+        (Δ′,)
+    end
+end
+
+Zygote.@adjoint function Base.map(f, args...)
```
vchuravy, line 44: I am surprised that we need a custom adjoint for map here

tools/zygote/hello.jl
athas: Yes. I will get rid of it.

tools/zygote/run_gmm.jl
athas: As am I. It was part of the original ADBench code, but it seems to work fine without it, so I will remove it.

Issue #96, opened 2024-10-22
Reach parity with ADBench


samestep:
Before we expand to supporting other benchmark programs, we should at least support what is already present in [ADBench](https://github.com/microsoft/ADBench). If I remember correctly, currently we're missing one or two of the ADBench evals, and while I think we support all of the tools, we have not yet migrated all of the evals for all of the tools. This is partially because we were previously planning to handle this "generally" by transpiling from [Adroit](https://adroit-lang.org/), which we have now decided not to do for now after some discussion with @athas.

athas:
Do we actually intend to port all of the ADBench tools? Some are obscure (like the ones for MATLAB) and many of the others are fragile or obsolete, and were never packaged in a reproducible way in the first place.

samestep:
@athas yeah if we decide there are some obscure ones from ADBench that we want to leave out, that's fine. What's the specific list you have in mind?

athas:
ADBench supports the following tools (in the ADBench paper, not sure how many more got added later):

* Adept (C++)
* ADIC2 (C++)
* ADOL-C (C++)
* Ceres (C++)
* clad (C++)
* Tapenade (C)
* DiffSharp (F#)
* ADiGator (MATLAB)
* ADiMat (MATLAB)
* MuPAD (MATLAB)
* ForwardDiff.jl (Julia)
* Autograd (Python)
* Theano (Python)

GradBench already has complete ports of Manual/Finite (which are not AD tools), and complete or partial ports of these tools:

* Tapenade
* DiffSharp
* Autograd

Additionally, GradBench has support for these that are not in ADBench (actually some are in the ADBench repository for sure):

* TensorFlow
* PyTorch
* Futhark
* Zygote
* SciLean
* Jax
* Enzyme

This leaves the following missing:

* Adept (C++): still seems actively maintained, and looks like it always worked well for ADBench. We should also have this.
* ADIC2 (C++): can't find any references newer than 2010, and no code, and the ADBench paper states they could not get it to work.
* ADOL-C (C++): still seems actively maintained, and looks like it always worked well for ADBench. We should also have this.
* Ceres Solver (C++): this is actively maintained, but it's not really an AD library, and it seemed to behave poorly for most of the ADBench benchmarks.
* clad (C++): actively maintained. It did not work properly for the ADBench authors (no support for arrays), but that was many years ago. Might be worth looking into.
* ADiGator (MATLAB): didn't work for the ADBench authors and seems unmaintained.
[...]

PR #495, opened 2025-04-15
Implement LSTM, BA, HT with Zygote

athas:
Based on ADBench. Unfortunately, `ba` and `ht` are not as easy to port (and I won't try), as they depend on `Zygote.forward`, which seems to not exist in newer versions of Zygote.


athas:
When this is merged, we can close #96.

PR #498, opened 2025-04-16
Add ForwardDiff.jl through DifferentiationInterface

vchuravy:
Since I feel it is my fault that #479 stalled out,
I wanted to open a PR with how I think we should proceed.

- Use DifferentiationInterface to implement "default" gradient variants in GradBench.jl
  - Currently I have only done "hello"
- Don't prescribe it's use, but leave it up to the person implementing support for a tool.

xref: #475

cc: @gdalle


gdalle:
@vchuravy I was planning on getting back to #479 no worries.

How is the current PR different from #479 though? To me the main difference seems to be that you didn't switch Zygote to DI. My approach did not force anything on the tool implementations from within GradBench either?

vchuravy:
The main difference is to not special-case the runner for DI. Basically, implementing my review comments :)

tools/zygote-jl/run.jl
athas, line 1: If we rename the directory, then the tool should also be renamed `zygote-jl` (the name is passed to `GradBench.main` in the `run.jl` file).

vchuravy:
>  but in the end there will be runner modifications anyway.

I was planning some runner modification to separate out data parsing and return data-wrangling, we can also add a "preparation stage".

gdalle:
> I was planning some runner modification to separate out data parsing and return data-wrangling, we can also add a "preparation stage".

That would be nice. Do you want to make it DI-agnostic as well? Cause that might be tricky.

gdalle:
I haven't had time to review yet, is it okay if I take a few days to think about how we go from @vchuravy's proposal to something that suits us both, in particular with respect to preparation? It seems we both represent the Julia community around these parts ^^

samestep:
> Also @samestep, do you have anything against renaming `zygote` to `zygote-jl`, as this PR does? Again, I have no preference. It does seem like most/all other Julia AD libraries have a `.jl` suffix.

Personally I'm not a huge fan, since it's not like there's a non-Julia version of Zygote, so I'd prefer it to be just `zygote`. But I'm not going to die on this hill if @vchuravy and/or @gdalle prefer it to be `zygote-jl` instead.

samestep:
> I haven't had time to review yet, is it okay if I take a few days to think about how we go from @vchuravy's proposal to something that suits us both, in particular with respect to preparation? It seems we both represent the Julia community around these parts ^^

@gdalle that sounds reasonable to me!

samestep:
Sure, that's fine.

gdalle:
Superseded by #575 which uses the new formalism with callable structs

PR #500, opened 2025-04-16
Add information on running outside of Docker

athas:


samestep (changes requested):
Thanks Troels! I'll be a bit more of a stickler on this one than usual, just because `README.md` is the most visible part of the project.

1. Instead of having a separate "Obtaining raw results" section, can you please just add the `-o` flag to the normal "Running GradBench locally" section? We should really be encouraging people to do this by default, I think; unless there's a specific reason not to.
2. Can you please move the "Without using Docker" section to be above the "Without cloning this repository" section? I think we probably both consider it to be much more important.
3. Personally I'd really rather not have a section on manually doing the Python stuff without uv via `pip`/`PYTHONPATH`/etc; but I could be convinced otherwise. Can you explain why that section is desirable? Has someone requested it, e.g. not being able to install uv themselves for some reason?
4. Please remove the `uv run` and `PYTHONPATH` stuff from the C++ section, as it is not necessary for running `python/gradbench/gradbench/cpp.py`.

athas:
>  Can you explain why that section is desirable? Has someone requested it, e.g. not being able to install uv themselves for some reason?

Max seemed to have trouble getting the right version of `uv`. I don't know if that is a common issue, or likely to be so in the future. It is very common for HPC installations to have tools that are a few years out of date.

> Please remove the uv run and PYTHONPATH stuff from the C++ section, as it is not necessary for running python/gradbench/gradbench/cpp.py.

This is true, but why would you ever run a tool without an eval? And the evals do require `uv` (or equivalent Python infraastructure).

samestep:
> Max seemed to have trouble getting the right version of `uv`. I don't know if that is a common issue, or likely to be so in the future. It is very common for HPC installations to have tools that are a few years out of date.

@MaxSagebaum to clarify, are you actually unable to install uv 0.6.8 or newer, or was it simply that you happened to have an older version at first when you tried running GradBench today?

> This is true, but why would you ever run a tool without an eval? And the evals do require `uv` (or equivalent Python infraastructure).

As I've mentioned before, I often run the eval in Docker while running the tool outside Docker. Also, perhaps in future we won't need uv to run the evals anyways, who knows.

athas:
> As I've mentioned before, I often run the eval in Docker while running the tool outside Docker. Also, perhaps in future we won't need uv to run the evals anyways, who knows.

Indeed, but not being able to run Docker *at all* is common in HPC environments, and I find it likely that many prospective GradBench users will be stuck with such environments. I want to be able to point them at an example of how to deal with that situation.

samestep:
Yeah, of course. I don't have any issue with giving instructions on how to run everything outside Docker, I just think that we shouldn't say that means you need to use uv when running the C++ tools specifically.

samestep (approved):
I've made some tweaks in aad9157079de3fe3811f2d1dd685b97bbeb7cccd:

- The name of the tool is usually written "uv", not "`uv`".
- In other places we don't prefix shell commands with `$ ` in Markdown, because GitHub provides a button to copy code blocks to the clipboard, and that doesn't work if there's an extra dollar sign. Ideally the dollar sign would render but not be included in copy-paste, but there's no way to do that in GitHub Markdown AFAIK.
- Fixed the link for installing uv.
- Used double quotes instead of single quotes for consistency.
- Replaced `uv run python/gradbench/gradbench/cpp.py` with `python3 python/gradbench/gradbench/cpp.py` in the second usage, for consistency (please correct me if this was actually intentional).
- Removed unnecessary `https://github.com/gradbench/gradbench/tree/main/` from a link.
- Sorted links at the bottom.
- Added `tools/manual` binaries to `.gitignore`.

I still really think we should put these built binaries in their own subfolders as I suggested in https://github.com/gradbench/gradbench/pull/473#issuecomment-2780813812, so that we don't have to keep maintaining all these `.gitignore` files.

@athas does this all look reasonable? I'm good to merge this if you are.

athas:
Looks good to me.

MaxSagebaum:
> > Max seemed to have trouble getting the right version of `uv`. I don't know if that is a common issue, or likely to be so in the future. It is very common for HPC installations to have tools that are a few years out of date.
> 
> @MaxSagebaum to clarify, are you actually unable to install uv 0.6.8 or newer, or was it simply that you happened to have an older version at first when you tried running GradBench today?
> 
> > This is true, but why would you ever run a tool without an eval? And the evals do require `uv` (or equivalent Python infraastructure).
> 
> As I've mentioned before, I often run the eval in Docker while running the tool outside Docker. Also, perhaps in future we won't need uv to run the evals anyways, who knows.

I just installed it via the package manager from Fedora.

PR #509, opened 2025-04-16
Fix #465

### LINKED ISSUE #465: `gradbench` sometimes does not propagate or discover errors
There have been a number of cases where `gradbench` runs just terminates with no output:

https://github.com/gradbench/gradbench/pull/463#issue-2968724319

Regardless of how we fix that issue, it is not great that the error message, which Docker *does* produce, is not propagated, or that the presumably nonzero exit code is not noticed.

athas:
This turned out to have a rather pedestrian cause: if no true build output is encountered, then the buffer is never printed.

Fixes #465.

### DIFF
--- a/crates/gradbench/src/main.rs
+++ b/crates/gradbench/src/main.rs
@@ -328,8 +328,11 @@ fn docker_build_quiet(color: Color, mut cmd: Command) -> anyhow::Result<ExitStat
             eprintln!("{}", line.color(color));
         }
     }
-    eprint!("{}", take(&mut buffer).color(color));
-    Ok(child.wait()?)
+    let status = child.wait()?;
+    if !status.success() {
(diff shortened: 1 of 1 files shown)


samestep (changes requested):
Sorry, this is not what we want; I specifically designed it this way in #347 so that it only prints the build output if it's not cached. I agree that we need to surface errors, but we need to find a way to do it that doesn't always print the entire build output every single time.

athas:
I thought I did test that case, but I suppose all of the Docker images are already cached here.

Perhaps we can print when `child.wait()` returns a failure?

athas:
I have done so.

samestep (approved):
Sweet, thanks! Looks good to merge once the Rust lint is fixed.

PR #512, opened 2025-04-16
Speed up gmm-zygote and ba-zygote

athas:
This is now more similar to ADBench. I also added an explanatory comment for why I think it is written this way.


athas:
Ugh, now gmm-zygote finishes 800ms before the timeout.

PR #513, opened 2025-04-16
Add `.clangd` and `.clang-format` files; Format many of the things.

aj-fleming:
Clangd is good. We should use it. 

Please note that `compile_flags.txt` has to exist to make clangd use relative paths, see: clangd/clangd/issues/1038

Clang-format is good. We should use it. The LLVM base style is nice, but I hold strong personal opinions about line breaks and like seeing consecutive using statements aligned neatly.

"wow, so nice"


.clang-format
```diff
@@ -0,0 +1,25 @@
...
+KeepEmptyLines:
+  AtEndOfFile: true
+  AtStartOfBlock: true
+  AtStartOfFile: false
+MaxEmptyLinesToKeep: 1
+
+NamespaceIndentation: Inner
+
+IncludeBlocks: Preserve
+
```
samestep, line 25: Does this file require `\n\n` at the end? If so then can we remove the stray blank line?

```suggestion
```

.clangd
```diff
@@ -0,0 +1,18 @@
...
+---
+# ad-hpp 
+If:
+  # use leading * to allow clangd to pick up the right files
+  # since clangd doesn't support relative paths to .clangd files
+  PathMatch: */tools/ad-hpp/*\.cpp
+
+CompileFlags:
+  Add: -Itools/ad-hpp/include
+
```
samestep, line 18: Same for this file.

```suggestion
```

samestep (commented):
Thanks! You've accomplished what I was too afraid to do.

I have a couple nits about stray newlines; other than that, I'm good to merge this if/when the `check-cpp` CI job passes.

samestep:
Huh, looks like Prettier has opinions about how to format `.clang-format` (see the `check-js` job log); could you fix that as well?

samestep (approved):
Alrighty, I think this is good to go. Thanks Sasha!

Issue #480, opened 2025-04-09
Preparation mechanism: is it fair game?


gdalle:
Some AD libraries benefit a lot from "preparation". Suppose we know in advance that we will compute a thousand gradients for a function `f` on inputs `x::Vector{Float64}` with a given size, and the only thing that changes is the value of `x`. Then, some preliminary computations can be amortized across all gradient evaluations because we only need to run them once. For instance, we can apply the source transformation, pre-allocate some memory for the tape, etc.
In Julia, [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl) provides a built-in [preparation mechanism](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/explanation/operators/#Preparation) which then specializes on a given AD backend.

When we know that an AD library can exploit preparation, would it be fair to run it in the first gradient evaluation and then reuse its products to speed up subsequent evaluations? In the current implementation for Julia (#479), as for most other libraries, gradient computations are independent from one another. But in realistic scenarios, like gradient descent, the input type and size does not vary, which means we can prepare once at the beginning and reap the results in the hot loop. Does it go against the spirit of GradBench?

athas:
This is generally allowed. Sam wrote his perspective the other day: https://github.com/gradbench/gradbench/pull/464#discussion_r2027321332

A tool can report multiple runtimes. All of the C++ runtimes also report a `"prepare"` runtime that covers things like preparing the data and preallocating tape and such. At a technical level, the code that does the preparation does have access to the input data, but using it would be against the spirit of things. It does however use the *size* of the input for the allocations.

Source transformations and such can also be applied in response to the `"define"` message. It is really up to the tool what goes where. The only thing that is iffy is moving dataset-sensitive work out of the code that is measured by the `"runtime"` measurements. It *is* however fine for the first measurement to prepare data (e.g. warm up caches) that are used by subsequent measurements in the measurement loop - they are not expected to be completely independent.

gdalle:
So if I understand correctly:

- preparations that use only the size and type can be excluded from benchmark measurements
- preparations that use actual inputs (like recording a tape when the number of iterations is part of the input) should be reported in the first run but can be removed from later run

I still don't understand the difference between `"prepare"` and `"define"` though

athas:
> preparations that use only the size and type can be excluded from benchmark measurements

[...]

PR #514, opened 2025-04-16
[Julia] define preprocess, execute, postprocess interface

vchuravy:
Currently the Julia interface is based around just functions, but we should move the `from_json` step out of the measurement loop and allow for post-processing as well. 

The syntax `(obj::Obj)(args...)` is used to create "callable" objects aka closures, but with the benefit that we can dispatch to `from_json()` and `postprocess` as well, but said syntax is also a bit "rarer" in usage. 

We could alternativly use:

```
function compute (::Experiment,args...)
end
```

But that is performance wise the same and just adds extra characters.

cc: @gdalle


athas:
So this is based on specialising generic functions? That seems like a good design to me. What is the purpose of `postprocess`?

vchuravy:
`postprocess` is in response to the comment.

>  It would be acceptable to move the massaging of the Jacobian
> into a separate function that is not timed, but I doubt it
>  matters much.

I'm case there is some processing needed to turn it from the natural answer to the expected answer.

athas:
Alright. In most cases that is the counterpart to `from_json` (which doesn't strictly speaking get JSON as input, but rather a dict), so maybe we could call it `to_json`.

gdalle (approved):
This looks good to me, up to potential renamings.
In this setting, I guess the preparation phase for DI could go in `to_json`? In which case, we could rename to `preprocess` to be more generic, and measure the corresponding time as discussed in #480?

athas:
I'm OK with a more general name if it needs to do more general things than massaging Jacobians.

julia/GradBench/src/GradBench.jl
```diff
@@ -2,48 +2,73 @@ module GradBench
...
 
 # Avoid measuring dispatch overhead
-function measure(func::F, arg) where {F}
+function measure(func::F, args...) where {F}
     start = time_ns()
-    ret = func(arg)
+    ret = func(args...)
     done = time_ns()
     return ret, done - start
 end
```
gdalle, line 41: This looks a bit clumsy now that we use it several times, can we just use `@timed`?

julia/GradBench/src/GradBench.jl
```diff
@@ -2,48 +2,73 @@ module GradBench
...
     timings = Any[]
 
+    args, t = measure(preprocess, experiment, params["input"])
+    push!(timings, Dict("name" => "preprocess", "nanoseconds" => t))
+
+    ret, t = measure(experiment, args...)
+    push!(timings, Dict("name" => "warmup", "nanoseconds" => t))
+
+    output, t = measure(postprocess, experiment, ret)
+    push!(timings, Dict("name" => "postprocess", "nanoseconds" => t))
```
gdalle, line 59: From a logical standpoint I would put this after the evaluation loop

julia/GradBench/src/GradBench.jl
```diff
@@ -2,48 +2,73 @@ module GradBench
...
     min_runs = get(params, "min_runs", 1)
     min_seconds = get(params, "min_seconds", 0)
     @assert min_runs > 0
 
-    # TODO: Prepare (parse JSON?)
-    # TODO: pre-allocate output?
     timings = Any[]
 
+    args, t = measure(preprocess, experiment, params["input"])
+    push!(timings, Dict("name" => "preprocess", "nanoseconds" => t))
```
gdalle, line 53: Judging by #480 this is called `"prepare"` for other tools?

julia/GradBench/src/GradBench.jl
```diff
@@ -2,48 +2,73 @@ module GradBench
...
+
+    ret, t = measure(experiment, args...)
+    push!(timings, Dict("name" => "warmup", "nanoseconds" => t))
```
gdalle, line 56: Judging by #480 this is meant to be measured inside the first execution?

julia/GradBench/src/GradBench.jl
```diff
@@ -2,48 +2,73 @@ module GradBench
...
+    experiment = mod[params["function"]]
     min_runs = get(params, "min_runs", 1)
     min_seconds = get(params, "min_seconds", 0)
     @assert min_runs > 0
 
-    # TODO: Prepare (parse JSON?)
-    # TODO: pre-allocate output?
     timings = Any[]
 
+    args, t = measure(preprocess, experiment, params["input"])
```
gdalle, line 52: The result of preparation/preprocessing is often type-unstable, for instance ForwardDiff picks a dynamic chunk size. Perhaps we could put a function barrier to shield the main execution loop from it?

gdalle (changes requested):

julia/GradBench/src/GradBench.jl
vchuravy: `experiment` is already type-unstable, which is why the `measure` is a function barrier and not a macro.

julia/GradBench/src/GradBench.jl
gdalle: Ok judging by https://github.com/gradbench/gradbench/pull/514#discussion_r2049253355 we want the function barrier, nevermind

julia/GradBench/src/GradBench.jl
athas: At least for the C++ tools, and there is no reason to diverge. In principle there is no limit to what fine-grained measurements a tool can provide (although the plots only care about `evaluate`).

gdalle:
Gentle bump on this one, since it would be a good basis for #498

athas:
I have merged `main` into this. `ode` with `enzyme-jl` works (just as it did before my merge), but the other implementations all need to be modified similarly.

athas:
I think I understand the design here, and if desirable, I can try to convert the remaining evals.

samestep:
> Gentle bump on this one, since it would be a good basis for #498

@gdalle I do also want to mention: I view it as a technicality that contributors cannot make edite to each others' pull requests, so if you'd like to make a separate branch/PR starting from this one and bring it closer to completion to unblock #498, you're more than welcome to do so! (Unless @vchuravy objects, of course.)

julia/GradBench/src/GradBench.jl
vchuravy: This running the code once, to do an exhaustive pre compilation.

If we just want top-level precompilation we could write

```
    ret, t = measure(precompile, (typeof(experiment),  map(typeof, args)...))
```

gdalle:
I can also complete this next week or so

athas:
That is not necessary. Barring oversights, they are all done now. It would be good however for someone fluent in Julia to take a look at my code.

gdalle:
I'll review next week

gdalle:
Sorry, the week was more busy than I thought, but it is still on my mind

vchuravy (commented):
Looks good to me! Thank you @athas

PR #525, opened 2025-04-18
julia: rewrite ode without side effects, use zygote

athas:
This PR rewrites the ode benchmark in a more array-oriented style (based on the Futhark implementation), such that Zygote can handle it. It is about x3 slower than the old imperative version, although I suppose that could eventually be recovered by parallelising the execution.

But there are questions here, which I think is best answered by the honoured representatives of the Julia community:

1) Is this a more idiomatic implementation than the previous imperative one?

2) How should we handle cases where one formulation of a benchmark is advantageous for one tool, but disadvantageous for others? (Or simply reduces primal performance.)

One possibility is to provide multiple implementations, and a tool can then pick the one that works the best. This is already done for the 'det' benchmark, which is a rather extreme case of this situation.

### DIFF
--- a/julia/GradBench/src/benchmarks/ode.jl
+++ b/julia/GradBench/src/benchmarks/ode.jl
@@ -9,6 +9,9 @@
 # module and an instantiation for the specific function under
 # consideration. In this implementation, this is all inlined for
 # implementation simplicity. This is follows the style use in "ode.hpp"
+#
+# This module provides two implementations: one that requires
+# mutation, and one that does not.
 
 module ODE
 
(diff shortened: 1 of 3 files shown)


athas:
@vchuravy @gdalle any opinions? I expect this will be a common issue when we add support for increasingly many Julia AD tools.

gdalle:
1. No, the mutating version is more idiomatic and we should keep both
2. We can just let the backend choose which primal function it differentiates

Your proposal is necessary for Zygote to work, but in general it will make the primal and dual computations _much much slower_ (unlike in Python where vectorized computations are always faster). Besides, with the advent of Enzyme and Mooncake in the Julia world, Zygote's "no mutation" constraint is increasingly looking like a thing of the past.

In general, I'd say we need to focus primarily on writing fast versions of every benchmark, even if they don't work with Zygote at first. Having mutation-free alternatives is great for comparison, but they should never replace the mutating implementations. Perhaps we could come up with a naming convention, like `ode_inplace` vs `ode_outofplace`?

athas:
For `det` I have two submodules, `Pure` and `Impure`.

athas:
Now I do the same here. The Impure module is used by Zygote.jl.

PR #529, opened 2025-04-18
cli: add 'log trim' command

athas:
This PR adds a new command to the gradbench CLI that sets the 'input' and 'output' fields of a log file to null. The purpose of this is to trim the often very large log files produced by GradBench, in the interest of saving on storage space and reading time. For example, the log file for gmm-futhark is reduced in size by about 95% after trimming. Most of the remaining space is taken up by timings.

The use of null allows us to distinguish trimmed log files, as presumably this value will never originally occur.

Although the command implementation is pretty short, I have factored it into a separate file 'log.rs', because I envision that we will eventually have a bunch of such log manipulation commands.

I am not a particularly experienced Rust programmer, so I would like feedback on my approach before I start adding more subcommands.

### DIFF
--- a/crates/gradbench/src/log.rs
+++ b/crates/gradbench/src/log.rs
@@ -41,3 +41,51 @@ pub fn trim(input: &mut impl BufRead, out: &mut impl Write) -> anyhow::Result<()
     }
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::log;
+    use goldenfile::Mint;
+    use std::io::{BufReader, Cursor, Write};
+
+    fn write_goldenfile(name: &str, bytes: &[u8]) {
+        let mut mint = Mint::new("src/outputs");
+        let mut file = mint.new_goldenfile(name).unwrap();
+        file.write_all(bytes).unwrap();
+    }
+
+    #[test]
+    fn test_trim_basic() -> anyhow::Result<()> {
+        let input = r#"{ "elapsed": { "nanoseconds": 3528846445 }, "message": {"id": 2, "kind": "evaluate", "module": "hello", "function": "square", "input": 1.0} }
+{ "elapsed": { "nanoseconds": 3543823169 }, "response": {"id": 2, "success": true, "output": 1.0, "timings": [{"name": "evaluate", "nanoseconds": 0}]} }
+"#;
+        let input_cursor = Cursor::new(input.as_bytes());
+        let mut output: Vec<u8> = Vec::new();
+        log::trim(&mut BufReader::new(input_cursor), &mut output)?;
+        write_goldenfile("trim_basic.jsonl", &output);
+        Ok(())
+    }
+
+    #[test]
+    fn test_trim_idempotence() -> anyhow::Result<()> {
+        let input = r#"{ "elapsed": { "nanoseconds": 3528846445 }, "message": {"id": 2, "kind": "evaluate", "module": "hello", "function": "square"} }
+{ "elapsed": { "nanoseconds": 3543823169 }, "response": {"id": 2, "success": true, "timings": [{"name": "evaluate", "nanoseconds": 0}]} }
+"#;
+        let input_cursor = Cursor::new(input.as_bytes());
+        let mut output: Vec<u8> = Vec::new();
+        log::trim(&mut BufReader::new(input_cursor), &mut output)?;
+        write_goldenfile("trim_idempotence.jsonl", &output);
+        Ok(())
+    }
+
+    #[test]
+    fn test_trim_missing_response() -> anyhow::Result<()> {
+        let input = r#"{ "elapsed": { "nanoseconds": 3528846445 }, "message": {"id": 2, "kind": "evaluate", "module": "hello", "function": "square"} }
+"#;
+        let input_cursor = Cursor::new(input.as_bytes());
+        let mut output: Vec<u8> = Vec::new();
+        log::trim(&mut BufReader::new(input_cursor), &mut output)?;
+        write_goldenfile("trim_missing_response.jsonl", &output);
+        Ok(())
+    }
+}
--- a/crates/gradbench/src/log.rs
+++ b/crates/gradbench/src/log.rs
@@ -1,0 +1,43 @@
+use crate::{
+    protocol::{EvaluateResponse, LogMessage, LogResponse, Message},
+    util::try_read_line,
+};
+
+use std::io::{BufRead, Write};
+
+pub fn trim(input: &mut impl BufRead, out: &mut impl Write) -> anyhow::Result<()> {
+    while let Some(line) = try_read_line(input)? {
+        let mut message: LogMessage = serde_json::from_str(&line)?;
+        match message.message {
+            Message::Evaluate {
+                id,
+                module,
+                function,
+                input: _,
+                description,
+            } => {
+                if let Some(response_line) = try_read_line(input)? {
+                    let mut response: LogResponse<EvaluateResponse> =
+                        serde_json::from_str(&response_line)?;
+                    message.message = Message::Evaluate {
+                        id,
+                        module,
+                        function,
+                        input: serde_json::Value::Null,
+                        description,
+                    };
+                    response.response.output = Some(serde_json::Value::Null);
+                    writeln!(out, "{}", serde_json::to_string(&message)?)?;
+                    writeln!(out, "{}", serde_json::to_string(&response)?)?;
+                }
+            }
+            _ => {
+                write!(out, "{}", line)?;
+                if let Some(response_line) = try_read_line(input)? {
+                    write!(out, "{}", response_line)?;
+                }
+            }
+        }
+    }
+    Ok(())
+}
--- a/crates/gradbench/src/log.rs
+++ b/crates/gradbench/src/log.rs
@@ -23,10 +23,10 @@ pub fn trim(input: &mut impl BufRead, out: &mut impl Write) -> anyhow::Result<()
                         id,
                         module,
                         function,
-                        input: serde_json::Value::Null,
+                        input: None,
                         description,
                     };
-                    response.response.output = Some(serde_json::Value::Null);
+                    response.response.output = None;
                     writeln!(out, "{}", serde_json::to_string(&message)?)?;
                     writeln!(out, "{}", serde_json::to_string(&response)?)?;
                 }
--- a/crates/gradbench/src/log.rs
+++ b/crates/gradbench/src/log.rs
@@ -29,6 +29,8 @@ pub fn trim(input: &mut impl BufRead, out: &mut impl Write) -> anyhow::Result<()
                     response.response.output = None;
                     writeln!(out, "{}", serde_json::to_string(&message)?)?;
                     writeln!(out, "{}", serde_json::to_string(&response)?)?;
+                } else {
+                    write!(out, "{}", line)?;
                 }
             }
             _ => {
--- a/crates/gradbench/src/main.rs
+++ b/crates/gradbench/src/main.rs
@@ -218,7 +218,7 @@ enum RepoCommands {
 
 #[derive(Debug, Subcommand)]
 enum LogCommands {
-    /// Change input/output fields of "evaluate" messages to null.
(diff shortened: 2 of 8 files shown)


samestep (changes requested):
Thanks!

I'll take a closer look at the code after lunch, but just from the PR description: I don't like replacing these values with `null`. I'd much rather just remove the fields entirely, since then it's crystal clear whether it's a trimmed log or a regular log, and there's no special cases where an eval or tool actually gave a `null` value. Can we do that instead?

athas:
If we remove `input`, then the log file is no longer valid according to the protocol, and would for example not be able to be deserialised again. We could fix that by making the `input` field optional - the `output` field is already optional, although I think that is intended for the error case.

samestep:
> If we remove `input`, then the log file is no longer valid according to the protocol, ...

Correct, it would no longer be valid according to the protocol; I think that's the right way to do it, because trimmed logs are a different kind of things from actual protocol-compliant logs. If we think it's useful then we could formalize that in TypeScript by defining a separate type for trimmed logs.

> ... and would for example not be able to be deserialised again. We could fix that by making the `input` field optional - the `output` field is already optional, although I think that is intended for the error case.

I don't want to make it optional in the actual protocol definition, because it's not optional; but if you want to make it optional just in our Rust type definitions so that we can share those same type definitions for future utilities using trimmed logs, that seems fine to me:

```diff
diff --git a/crates/gradbench/src/protocol.rs b/crates/gradbench/src/protocol.rs
index 5e61f413..2178c2e9 100644
--- a/crates/gradbench/src/protocol.rs
+++ b/crates/gradbench/src/protocol.rs
@@ -45,7 +45,7 @@ pub enum Message {
         function: String,
 
         /// The input to the function.
-        input: serde_json::Value,
+        input: Option<serde_json::Value>,
 
         /// A short human-readable description of the input.
         description: Option<String>,
@@ -118,7 +118,7 @@ pub struct EvaluateResponse {
         deserialize_with = "deserialize_optional_json", // Deserialize as `Some` if present.
         skip_serializing_if = "Option::is_none" // Serialize as missing if `None`.
     )]
-    pub output: Option<serde_json::Value>,
+    pub output: Option<Option<serde_json::Value>>,
 
     /// Subtask timings.
     pub timings: Option<Vec<Timing>>,
```

athas:
I am OK with those trimmed logs, since they are superset of the normal log format, and I don't expect I will ever process them in anything but Rust or Python. I don't think it is worth documenting it as a separate format - it is likely to just confuse users, and they really don't need to know the distinction.

athas:
Alright, that is done. It is my now my impression that this is good code.

Another, smaller-stakes command I want to do is `gradbench log summary`, which given a log file tells you some basic information about the eval and tool, their configuration, total runtime, etc.

I will also eventually get around to `gradbench log tabulate`, which will extract tabular runtime data for ease of plotting etc.

crates/gradbench/src/main.rs
```diff
@@ -119,6 +120,12 @@ enum Commands {
         #[command(subcommand)]
         command: RepoCommands,
     },
+
+    /// Perform useful operations on the log files produced by `gradbench run`.
+    Log {
+        #[command(subcommand)]
+        command: LogCommands,
+    },
```
samestep, line 128: A bit nitpicky, but can you move this above the `Repo` variant? I want to keep `Repo` at the bottom since it's the only category that really has different semantics from all the rest (it checks to see if there's a `.git` folder in the current working directory).

crates/gradbench/src/main.rs
```diff
@@ -675,6 +701,18 @@ fn cli() -> Result<(), ExitCode> {
...
+                let input_file = fs::File::open(input).map_err(|err| err_fail(anyhow!(err)))?;
+                let mut output_file: Box<dyn std::io::Write> = match output {
+                    Some(path) => {
+                        Box::new(fs::File::create(&path).map_err(|err| err_fail(anyhow!(err)))?)
+                    }
+                    None => Box::new(io::stdout()),
+                };
+                log::trim(&mut io::BufReader::new(input_file), &mut output_file).map_err(err_fail)
+            }
+        },
```
samestep, line 715: The repeated use of `anyhow!` and `err_fail` here adds a lot of noise. The reason you found yourself having to do this is because this function returns `Result<(), ExitCode>` instead of `anyhow::Result<()>`, so you aren't really able to make effective use of the `?` operator. Could you please instead move this into a separate function that returns `anyhow::Result<()>` and just call that, then have one single usage of `.map_err(err_fail)` here?

crates/gradbench/src/main.rs
```diff
@@ -675,6 +701,18 @@ fn cli() -> Result<(), ExitCode> {
                 }
             }
         }
+        Commands::Log { command } => match command {
+            LogCommands::Trim { input, output } => {
+                let input_file = fs::File::open(input).map_err(|err| err_fail(anyhow!(err)))?;
+                let mut output_file: Box<dyn std::io::Write> = match output {
```
samestep, line 707: Instead of using `dyn` here, could you just put separate calls to `log::trim` in the `Some` and `None` branches, like we do for `Commands::Run`? That way we avoid the unnecessary dynamic dispatch.

crates/gradbench/src/log.rs
samestep, line 1: Could you add a unit test in this file?

samestep (changes requested):
Thanks, this looks pretty good overall! Just a few requests.

crates/gradbench/src/main.rs
athas: For consistency with Unix conventions and ease of pipelining, I would also like to make the input option optional. Does this mean I should have four control paths? And if I understand the Rust compilation model, also four duplicated monomorphised functions?

samestep: Correct.

athas:
I have made the requested changes. I don't understand why you prefer the approach with duplicated calls to `log::trim`, however.

samestep:
By the way, I forgot to mention this, but for unit testing, I strongly recommend just using the goldenfile approach: just make an input `*.jsonl` file and then in the test, run the function and use the goldenfile crate to write to an output `*.jsonl` file.

athas:
I'll try to write a few tests now.

samestep:
I haven't looked at the code changes yet though (currently on my phone).

samestep:
> I have made the requested changes.

Thank you!

> I don't understand why you prefer the approach with duplicated calls to `log::trim`, however.

Just because I think it's a good default to not pay the dynamic dispatch cost without a specific reason, and any (source or generated) code duplication cost is minimal.

athas:
These are IO functions; you won't even be able to measure the dynamic dispatch cost compared to everything else they do. The VFS layer of every kernel implements file objects as a pair of a pointer and a vtable, which is exactly what a `dyn` is. There will be a dynamic dispatch cost because that is how everyone implements a file abstraction.

I worry the cost in duplication will be paid later, when the command grows more options and the duplicated function call becomes longer.

athas:
Is there a good way to store the input files for `goldenfile`, or do you suggest I keep them as raw string constants in the source?

athas:
Alright, I added unit tests. Since the inputs are relatively small, I put them in as string constants.

crates/gradbench/src/main.rs
```diff
@@ -675,6 +728,7 @@ fn cli() -> Result<(), ExitCode> {
                 }
             }
         }
+        Commands::Log { command } => log_command(command).map_err(|err| err_fail(anyhow!(err))),
```
samestep, line 731: No need to wrap this in another `anyhow!`:

```suggestion
        Commands::Log { command } => log_command(command).map_err(err_fail),
```

crates/gradbench/src/main.rs
```diff
@@ -567,6 +593,33 @@ fn matrix() -> anyhow::Result<()> {
     Ok(())
 }
 
+fn log_command(command: LogCommands) -> anyhow::Result<()> {
```
samestep, line 596: Can you put a docstring on this function, just for consistency with the rest of the code in this file? It can be short.

crates/gradbench/src/log.rs
samestep, line 1: Looks like you forgot to commit the `*.jsonl` output files; you can generate those automatically by just running this command:

```sh
UPDATE_GOLDENFILES=1 cargo test
```

But also, I realize that I forgot to write down something important in my earlier review. In these outputs, the `"description"` and `"error"` fields come out as `null`, which is invalid; those fields should either be missing or be strings. Can you fix this aspect of the serialization?

samestep (changes requested):
> Is there a good way to store the input files for `goldenfile`, or do you suggest I keep them as raw string constants in the source?

I typically like making them separate files just so that they're easier to read in a text editor (you get syntax highlighting that way).

> Alright, I added unit tests. Since the inputs are relatively small, I put them in as string constants.

Sure, that's fine.

---

> These are IO functions; you won't even be able to measure the dynamic dispatch cost compared to everything else they do. The VFS layer of every kernel implements file objects as a pair of a pointer and a vtable, which is exactly what a `dyn` is. There will be a dynamic dispatch cost because that is how everyone implements a file abstraction.

That makes sense, thanks for explaining; I'm not as familiar with those details, so I apologize for my ignorance. If you want to change it back to `dyn` in this case, that's fine. I have more thoughts on the general case (beyond this specific function), but that seems better suited for a separate discussion.

> I worry the cost in duplication will be paid later, when the command grows more options and the duplicated function call becomes longer.

What other options do you foresee being added that would actually result in different types being passed?

athas:
> What other options do you foresee being added that would actually result in different types being passed?

None, but if we add another five (non-file-related) options, then each of the four calls to `trim` end up with an additional five arguments. It just starts getting quite large. I can easily imagine adding options that control what to trim (e.g. maybe someone wants to replace the input/outputs with hashes instead, or also trim the runtimes, or something appropriate to weird evals we don't yet have any conception of).

samestep:
> None, but if we add another five (non-file-related) options, then each of the four calls to `trim` end up with an additional five arguments. It just starts getting quite large. I can easily imagine adding options that control what to trim (e.g. maybe someone wants to replace the input/outputs with hashes instead, or also trim the runtimes, or something appropriate to weird evals we don't yet have any conception of).

I think in this case the right approach is to package arguments into a struct:

https://github.com/gradbench/gradbench/blob/db31af4fc0ff1c93e5521a417d7d27c52daeb336/crates/gradbench/src/main.rs#L691-L702

But it is unfortunate that Rust's approach for doing that is so clunky. I wish Rust had anonymous struct types like it has anonymous tuple types.

crates/gradbench/src/log.rs
athas: I did so, and also fixed the other optional fields while I was at it.

crates/gradbench/src/log.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    #[test]
+    fn test_trim_missing_response() -> anyhow::Result<()> {
+        let input = r#"{ "elapsed": { "nanoseconds": 3528846445 }, "message": {"id": 2, "kind": "evaluate", "module": "hello", "function": "square"} }
+"#;
+        let input_cursor = Cursor::new(input.as_bytes());
+        let mut output: Vec<u8> = Vec::new();
+        log::trim(&mut BufReader::new(input_cursor), &mut output)?;
+        write_goldenfile("trim_missing_response.jsonl", &output);
```
samestep, line 88: It looks like, when the file is malformed, the function just fails silently with no output rather than returning an error. That doesn't seem like the desired behavior, right? Shouldn't it either give a nonempty output here, or an error?

samestep (commented):
Thanks for addressing everything! Just had one more question about one of your test cases.

crates/gradbench/src/log.rs
athas: I had a different semantics in mind, which I documented but forgot to implement.  Now it works. A log with a missing response is not malformed - it can be produced by the CLI due to timeouts or crashes.

samestep (approved):
Looks great! Thanks for implementing this and being patient with me.

PR #531, opened 2025-04-19
shell.nix: add rust-analyzer

athas:


samestep:
Ha, funny; never noticed this was missing because I just use the one that comes bundled with [the VS Code extension](https://marketplace.visualstudio.com/items?itemName=rust-lang.rust-analyzer). But I guess you need it for Emacs yeah?

athas:
Yes.

PR #533, opened 2025-04-19
cli: add 'log summary' command

athas:
This prints a summary of pertinent information in log files. It might look like this:

```
$ ./gradbench log summary gmm-futhark.trimmed.jsonl
eval: gmm
config: {"d":[2,10,20,32,64],"k":[5,10,25,50,100],"min_runs":1,"min_seconds":1,"n":1000,"no_validation":false}
tool: futhark
config: {"backend":"c"}
evaluations: 50
      valid: 50
    invalid: 0
elapsed:  1:01.424
```

This can be quite useful when you've got a whole bunch of log files and forgotten wha they correspond to.

The PR looks large because I had to add a bunch of protocol fields that we were missing. I also moved nanostring into the util module so I could reuse its string formatting machinery without importing the intermediary. All else equal, I think there is value in the CLI being consistent in how it prints times.


athas:
Oh, this doesn't have any of today's mandatory fruit salad output yet. Any suggestions for what colours to put in?

crates/gradbench/src/intermediary.rs
samestep, line 1: I'm thinking I should probably refactor the tests in this file so that they don't have to write `None` for all the omitted fields all the time, it's annoying to have to constantly edit these; maybe the inputs should just be factored out into `*.jsonl` files.

No need to worry about that in this PR, just thinking out loud.

crates/gradbench/src/main.rs
```diff
@@ -1054,6 +1064,16 @@ fn log_command(command: LogCommands) -> anyhow::Result<()> {
...
+        LogCommands::Summary { input } => match input {
+            Some(input_path) => {
+                log::summary(&mut io::BufReader::new(fs::File::open(input_path)?))?;
+                Ok(())
+            }
+            None => {
+                log::summary(&mut io::BufReader::new(io::stdin()))?;
+                Ok(())
+            }
+        },
```
samestep, line 1076: I'm guessing in the future we could make a utility function that does this one-time dispatching automatically. Or, alternatively, possibly a better design could be to have the `summary` function just take an `Iterator` of log lines. Again, no need to change it in this PR, I'm just thinking about stuff.

crates/gradbench/src/protocol.rs
samestep, line 1: Another pondering: perhaps we should make a separate type that has the serialization behavior that we want, instead of duplicating this same Serde annotation onto every usage of `Option` in this file.

samestep (changes requested):
Looks good, thanks! I wrote a few comments but none of them are blockers.

One blocker, though: can you please write at least one unit test? Like before, goldenfile is gonna be the easiest way; probably what you wanna do is refactor the `summary` function so it takes an `impl Write` or something, so in the test you can just have that be the file.

athas:
Tests are for the insufficiently confident, but I have now added some for `summary`.

samestep (changes requested):
Sweet thx! Can you rename the test output files from `*.jsonl` to `*.txt` though?

samestep (approved):
Thank you!

PR #538, opened 2025-04-22
llsq: strength reduction

athas:
This replaces the pow with an accumulator, like in the Floretta implementation. The speedup is pretty dramatic.

The Futhark tool probably fails, as it depends on a compiler version that isn't going to be released until about 30 minutes from now.


samestep (approved):
The blue square was too good to last...

PR #542, opened 2025-04-23
cpp: add common.mk

athas:
This adds a common set of Makefile rules that can be used by all C++ tools. This makes it easy to ensure a consistent set of optimisation flags; e.g. using -march-native.

There is no requirement for a tool to use common.mk, or these optimisation flags, if some specialised set happens to work better for some tool. The purpose is just to avoid differences due to arbitrary differences in how the compiler is configured.

(In this regard, let us ignore that enzyme uses a completely different compiler than all other C++ tools...)


samestep (approved):
Looks cool to me!

PR #544, opened 2025-04-23
ht-tapenade: actually exploit sparsity

athas:
This makes it orders of magnitude faster and also use orders of magnitude less memory somehow. The fact that this saves memory probably implies a memory leak in hand_objective_complicated_d().


athas:
I've written this program almost a dozen times by now. Waste of neurons.

PR #551, opened 2025-04-24
add cref to function wrapper object in `gd.hpp`

aj-fleming:
- this was the likely source of some gnarly bugs (read: I didn't implement rule-of-three) keeping track of tape pointers.

I suspect that, since the argmax function takes f by const-reference, this change should be ok.

### DIFF
--- a/cpp/gradbench/gd.hpp
+++ b/cpp/gradbench/gd.hpp
@@ -37,8 +37,18 @@ T vector_dist(const std::vector<T>& u, const std::vector<T>& v) {
   return sqrt(acc);
 }
 
+/**
+ * @brief Finds the multivariate argmin of a function via gradient descent.
+ * @tparam F A function type that has a member functions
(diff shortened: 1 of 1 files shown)


athas:
Rule-of-three? Why do discussions about C++ always sound like theological debates where problems are the result of moral failings?

aj-fleming:
To keep track of the allocated tape memory and prevent spurious allocations, I don't want to pass new [adjoint drivers](https://github.com/gradbench/gradbench/blob/0119df6ac0ebfb01552d84afadf8e11d6f47b21d/tools/ad-hpp/particle.cpp#L40) every time [compute](https://github.com/gradbench/gradbench/blob/0119df6ac0ebfb01552d84afadf8e11d6f47b21d/tools/ad-hpp/particle.cpp#L257) is called... replacing `const F` with `const F&` prevents calls to the copy constructor of the drivers.

PR #557, opened 2025-04-25
Refactor input/output polymorphism for `log trim`

samestep:
To address the concerns of code duplication from #529, this PR creates an `InOut` trait with a helper function that should make it easy to handle this pattern in the future, e.g. for #533.

### DIFF
--- a/crates/gradbench/src/log.rs
+++ b/crates/gradbench/src/log.rs
@@ -1,54 +1,59 @@
 use crate::{
     protocol::{EvaluateResponse, LogMessage, LogResponse, Message},
-    util::try_read_line,
+    util::{try_read_line, InOut},
 };
 
-use std::io::{BufRead, Write};
+use std::io;
 
-pub fn trim(input: &mut impl BufRead, out: &mut impl Write) -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


samestep:
If you feel strongly about not requiring a hyphen then I can undo the semantic change, and have this PR just be a refactoring to introduce the `InOut` trait. So, separately from the semantic question, does the refactoring look OK?

However, I do also want to push back a little bit. Is it actually that annoying to have to type a hyphen here?

athas:
I am unopposed to the `InOut` change. I find it a bit strange, but I suppose may be idiomatic in Rust.

And yes, requiring a hyphen in a pipeline is non-idiomatic. I am not aware of where the hyphen convention came from, but I am opposed to it - `/dev/stdin` exists and is clearer and requires no special code in programs. My experience is that the hyphen convention is mostly used by tools that are *typically* not used in pipelines (e.g. the input to a C compiler), and my guess is that it predates the existence of `/dev/stdin`.

samestep:
@athas I've made the change; does this look OK now?

athas (approved):

PR #558, opened 2025-04-26
Make C++ tools put binaries in bin/

### LINKED ISSUE #515: C++ `.gitignore` files are hard to maintain
See https://github.com/gradbench/gradbench/pull/473#issuecomment-2780813812 and https://github.com/gradbench/gradbench/pull/500#pullrequestreview-2772863072. Currently we have `tools/*/.gitignore` for many C++ tools, and each one just lists the names of some subset of the evals. These files are often incomplete; for instance:

```sh
make -C tools/ad-hpp
```

After running the above command, the files `tools/ad-hpp/hello` and `tools/ad-hpp/llsq` exist but are not in `.gitignore`.

I propose that, instead of littering these binaries around, we put them in `tools/*/bin` or `tools/*/out` folders.

athas:
This makes it easier to write and maintain a general .gitignore.

The change is mostly transparent to users. The one exception is deviants who compile C++ directly; they will need to use

    make bin/foo

instead of merely

    make foo

Closes #515.


samestep (approved):
Thanks so much! In 8a0fbe04522915d11232024cf988697374620e80 I fixed a couple things that got accidentally removed by 846617c01886933d9fd336a15e6d071692ac28fb, split the root `.gitignore` into more commented sections, and also removed some outdated entries that we no longer need.

samestep:
Looks like a couple of the `adol-c` runs are failing in CI, though?

Issue #390, opened 2025-03-19
TensorFlow is expected to error on HT but sometimes times out instead


samestep:
[Timeout example:](https://github.com/gradbench/gradbench/actions/runs/13935681447/job/39004890007)

```
 [38] eval  ht::objective   hand10_t26_c... 10:01.561   ⧖
expected exit code 1 (from outcome error) but got exit code 5
```

[Non-timeout example:](https://github.com/gradbench/gradbench/actions/runs/13932892445/job/38995679510)

```
 [36] eval  ht::jacobian    hand9_t26_c1...  5:07.009  

invalid JSON from tool: EOF while parsing a value at line 1 column 0
```

This makes CI flaky, and the flakiness can even be seen in our commit history: #287 originally switched TensorFlow's expected outcome on `ht` from `success` to `timeout`, and then the next day, #371 switched it again from `timeout` to `error`.

athas:
Yeah, this sucks. I'm tempted to just change the timeout to five minutes instead. Fixed forever!

athas:
Sure, if you're up for implementing it! I'm a bit averse to introducing complexity for the benefit of ornery tools that can't even manage to fail in a consistent manner.

samestep:
Or actually, it was more with the goal of being able to support less-mainstream tools like `haskell`, right? Since we wanted to be able to plot the data for the parts of the eval the tool _could_ complete, without requiring it to finish the entire eval to get any data at all.

athas:
If you're feeling generous with your time, I think the best solution is for a tool to declare which workloads it is willing to try, with both timeout and failure then always treated as real problems. Hopefully that would not be flaky. But I don't really have a proposal for a design for it. The current design was more of an expedient compromise (which frankly is still in its current state far better than most other benchmark suite).

samestep:
Yeah, personally I'm not really in favor of that since it'd be giving more semantic meaning to the `"description"` field than I'd want to. I think supporting a set of possible outcomes should cover our needs. 🤞

athas:
It is also interesting that the niche academic language Haskell is not the troublemaker here, but rather that the grown-up adult serious industry tool Tensorflow is not deterministic in whether it times out or runs out of memory. Clearly Haskell has a way to go before it is ready for industry.

athas:
Oh, but the workload selection can be done without considering the description. If we assume workloads are increasing in size, then the tool can declare willingness to handle the first N ones.

PR #560, opened 2025-04-30
ht-tensorflow: error out on large workloads

athas:
This is to work around the CI flakiness documented in #390.

If the implementation is ever improved, this bit of sabotage is easy enough to remove.


samestep (approved):
Makes sense to me; thanks!

Issue #373, opened 2025-03-13
Timeouts hide invalid outputs


samestep:
For instance, on commit c918e2e0b9d2969b5c3bf8dd65501ed5d965b159:

```
$ ./gradbench run --eval "./gradbench repo eval gmm" --tool "./gradbench repo tool finite" --timeout 120
  [0] start gmm (finite)
  [1] def   gmm                                 3.089 s ✓
  [2] eval  gmm::objective  2_5_1000            1.044 s ~         0ms prepare,         0ms evaluate × 19413 ✓
  [4] eval  gmm::jacobian   2_5_1000            1.010 s ~         0ms prepare,         3ms evaluate × 329 ✓
  [6] eval  gmm::objective  2_10_1000           1.031 s ~         0ms prepare,         0ms evaluate × 12737 ✓
  [8] eval  gmm::jacobian   2_10_1000           1.011 s ~         0ms prepare,         9ms evaluate × 107 ✓
 [10] eval  gmm::objective  2_25_1000           1.018 s ~         0ms prepare,         0ms evaluate × 5717 ✓
 [12] eval  gmm::jacobian   2_25_1000           1.011 s ~         0ms prepare,        52ms evaluate × 19 ✓
 [14] eval  gmm::objective  10_5_1000           1.025 s ~         0ms prepare,         0ms evaluate × 2095 ✓
 [16] eval  gmm::jacobian   10_5_1000           1.280 s ~         0ms prepare,       315ms evaluate × 4 ✓
 [18] eval  gmm::objective  2_50_1000           1.012 s ~         0ms prepare,         0ms evaluate × 3069 ✓
 [20] eval  gmm::jacobian   2_50_1000           1.196 s ~         0ms prepare,       198ms evaluate × 6 ✓
 [22] eval  gmm::objective  10_10_1000          1.024 s ~         0ms prepare,         0ms evaluate × 1048 ✓
 [24] eval  gmm::jacobian   10_10_1000          1.268 s ~         0ms prepare,     1.249 s evaluate ✓
 [26] eval  gmm::objective  20_5_1000           1.035 s ~         0ms prepare,         0ms evaluate × 1015 ✓
 [28] eval  gmm::jacobian   20_5_1000           2.173 s ~         0ms prepare,     2.131 s evaluate ✓
 [30] eval  gmm::objective  32_5_1000           1.052 s ~         0ms prepare,         1ms evaluate × 651 ✓
[...]

PR #561, opened 2025-05-06
Add TensorFlow.js as a tool

samestep:
This PR implements our `hello`, `llsq`, and `lse` evals with [TensorFlow.js](https://www.tensorflow.org/js).


samestep:
I attempted this implementation for `js/tensorflow/lse.ts`:

```typescript
import type { Runs } from "@gradbench/common/protocol.ts";
import { multipleRuns } from "@gradbench/common/util.ts";
import * as tf from "@tensorflow/tfjs";

const f = (x: tf.Tensor): tf.Tensor => {
  const a = x.max();
  return a.add(x.sub(a).sum().log());
};

interface Input extends Runs {
  x: number[];
}

export const primal = multipleRuns(({ x }: Input) => {
  return () => f(tf.tensor(x)).arraySync();
});

export const gradient = multipleRuns(({ x }: Input) => {
  const g = tf.grad(f);
  return () => g(tf.tensor(x)).arraySync();
});
```

But it failed validation because [apparently TensorFlow.js doesn't support 64-bit floating point](https://stackoverflow.com/a/57425824/5044950):

```
$ ./gradbench repo run --eval lse --tool tensorflow-js
building eval lse

building tool tensorflow-js

running eval lse
   with tool tensorflow-js
  [0] start lse
  [1] def   lse                                   170ms ✓
  [2] eval  lse::primal     n=2500         
============================
Hi, looks like you are running TensorFlow.js in Node.js. To speed things up dramatically, install our node backend, visit https://github.com/tensorflow/tfjs-node for more details. 
============================
     1.007 s ~         0ms evaluate × 10651
  [3] eval  lse::gradient   n=2500              1.005 s ~         1ms evaluate × 878 ✗
Found 2500 mismatches, showing 10:
[0]: 0.0004779547863593653 != -0.000804668408818543
[1]: 0.0006103941100944994 != -0.000804668408818543
[2]: 0.000310043651920869 != -0.000804668408818543
[3]: 0.000275616459044466 != -0.000804668408818543
[4]: 0.0003367300656552756 != -0.000804668408818543
[5]: 0.0006189169165366948 != -0.000804668408818543
[6]: 0.000555053736366889 != -0.000804668408818543
[7]: 0.0002721678220624913 != -0.000804668408818543
[8]: 0.00028903946470786957 != -0.000804668408818543
[9]: 0.0004640129292467529 != -0.000804668408818543
  [5] eval  lse::primal     n=5000         ^C
       517ms                                                                                                                                       

outcome interrupt
```

samestep:
I'm reminded that I should probably deal with #373.

samestep:
@athas so just to clarify, are you saying we shouldn't merge this PR because there's no way to make it use 64-bit floating point?

athas:
I think merging it is fine as long as the README makes it clear that the runtime numbers are not comparable with other tools, because it uses the wrong precision. Validation might also become a problem in some cases - I haven't ever really pondered the numerical stability of our evals.

samestep:
Hmm... if the numbers produced wouldn't be comparable with other tools then it seems like a bad idea to merge, since anything merged into this repo gets included in the autogenerated charts.

I'm thinking it would be nice to be able to select between single and double precision in the same way that we're adding the ability to select between single-threaded and multi-threaded. Thoughts? (Also this feels like it might be a slightly longer conversation, so if you'd prefer to do it in a separate GitHub issue and/or on Discord instead, that's fine with me.)

athas:
For many of the tools, such a toggle would be awkward to implement. For all of them, it would make the code more complicated. Some of the tools might not support single precision at all. I do not consider it worth it, as precision has no impact on AD - it only affects details about code generation. Barring cases where the choice of precision has some dramatic but arbitrary code generation effect (e.g. maybe some accelerated instruction for inverse square roots is only available for single precision), I would expect single precision variants to be 1.5x-2x faster on larger workloads than double precision, simply because they take up half as much space in the cache. I would expect the AD overhead to be the same. I suppose it is a worthwhile experiment to verify whether this is indeed the case, but I am skeptical of whether that question is interesting enough to justify the implementation effort. It seems better to just take the objective runtimes of tensorflow-js with a grain of salt.

samestep:
@athas the more I think about it, the more I actually do want to make a `--precision` CLI argument for tools. I think it's totally fine if some of the tools don't support it at all; I don't really see a reason to prioritize those over tools that don't support double precision. Especially since you mentioned that the fields of graphics and machine learning particularly like single-precision; I don't think we can really call GradBench "a benchmark suite for differentiable programming across languages _and domains_" if we're specifically excluding those two major fields.

I'm going to start working on a `--precision` PR, unless you have stronger objections than you've written thus far.

samestep:
@athas as I've implemented here in 3427b1ea06ff9ceca120ad4c865aaabaeab234c8, my intention is that if a tool doesn't support the level of precision requested (with `--precision=double` being the default, as it has been so far), it simply returns an error in response to a `"define"` message.

samestep:
Then when we get around to replacing `evals.txt` with `gradbench.json`, we can let tools specify information about what precision(s) they support, along with other information we really should be gathering there, like what color should be used for that tool in plots.

athas:
>  I don't think we can really call GradBench "a benchmark suite for differentiable programming across languages and domains" if we're specifically excluding those two major fields.

Support for such fields is orthogonal to whether we support a `--precision` argument. An eval can just specify that the computation must (or may) be done in single precision. It might actually be inappropriate to do neural networks with double precision.

A `--precision` argument also supposes a *single* precision, but the interesting case (and an increasingly common case for neural networks) is *mixed precision*.

(But I'm still not sure it has any nontrivial impact on AD.)

samestep:
Marking this as draft for now since what I really want is to use the WebGL backend which is the default one for the browser; apparently I'd have to do that via something like [Puppeteer](https://pptr.dev/).

athas:
> So just to clarify: are you saying that all our current evals depend on double precision to such a degree that it would simply be meaningless to use single-precision for them?

I don't know about "meaningless". It is certainly un-idiomatic within some of the fields from which those evals are derived (particularly classic scientific computing) to use anything but double precision. I know that the ADBench-derived evals will probably validate with single precision, because I have used such variants for research in the past (to run on machines with poor support for double precision). This is why I say that the effort is not worth it.

> I hadn't really seen mixed-precision used; could you give a link? (I'm curious to learn more.)

I'm not an ML person (at least not that kind), so I mostly know of it from the performance side of things. It is famously used by NVIDIA for tensor cores: https://docs.nvidia.com/deeplearning/performance/mixed-precision-training/index.html

> I guess I don't understand why we can't also have a toggle to configure other tools to use single precision. It's pretty easy for JAX as shown in https://github.com/gradbench/gradbench/pull/562, and I suspect it wouldn't be too hard for the C++ tools either, using similar machinery to what you've already built to support multithreading.

It is a tradeoff. The changes to support multithreading made the code more difficult to read and maintain, but I think that is justified because multithreadeding is obviously an important aspect of AD. Toggling precision *for the current set of evals* is not particularly important, and the benefit is not commensurate with the cost. The Jax and PyTorch tools are special cases where the cost is low. For the C++ implementations, the cost is rewriting to make the code more polymorphic. Currently the implementations use plain `double` for those values that are known to be inactive, and are only templatized for the active values (this is needed for the tape-based systems anyway). But it would not be truly *difficult* to rewrite these programs to support arbitrary (but fixed) precision, as I assume all the AD tools support it. It would make the scaffolding/driver code more complicated too, however.

Some of the other languages are more tricky. OCaml straight up does not support single precision at all. Haskell does support it, but I think the numerically optimised parts of `ad` are only for double precision. Futhark has excellent support for everything, but you need to rewrite the code to be generic (similar to what is needed for C++, although with much better error messages). Julia support `Float32`, but I think most literals and such will be `Float64`, and it might be awkward to write precision-polymorphic code. But you'd probably need to ask a Julia expert to be sure. How do you propose it would be done for Floretta, where WASM has no ability to abstract over types at all?

What is clear is that such polymorphic code will look more complicated and non-idiomatic than monomorphic code. While readability of the benchmarks is not a stated goal of GradBench, I do find it a valuable side benefit that our evals are also readable examples of how to use various tools. It's a benefit that can be sacrificed when there is a good reason to do so, of course.

I'm not opposed to playing with precision *in principle*, and I have done so in my own benchmarks previously. Sometimes it can reveal interesting things. Were the cost zero, I would certainly support it, but I do not think the cost is zero, and it would be more of a *tax* than a one-time cost, as this polymorphic code would have to be maintained. I am not going to argue very strongly against it, but I am unlikely to be the one who makes the very tedious modifications of the existing evals to suport this.

samestep:
@athas thanks, this makes sense. Given that reasoning, I'm fine with leaving alone all the tools you mentioned for which it'd be difficult to allow switching between single and double precision.

For this PR specifically, to me it actually seems fine to just merge TensorFlow.js into GradBench with no special treatment, since I realized that I just had a stupid bug in my code earlier and it actually validates just fine for the `llsq` and `lse` evals. Does that seem reasonable to you? Then when/if I get around to implementing other evals like `gmm` for which tools don't seem to be able to validate with single precision, this should be fine as long as we display the fact that it doesn't validate by coloring that cell red or something.

Then for #562, do you think it'd be reasonable for me to just add that flag for the tools that make it easy (like JAX and I assume PyTorch and/or TensorFlow), and then just merge it, leaving the others alone?

athas:
> For this PR specifically, to me it actually seems fine to just merge TensorFlow.js into GradBench with no special treatment

I agree, but the README should make it clear that the implementation uses single precision, just like we document other implementation quirks.

> Then for https://github.com/gradbench/gradbench/pull/562, do you think it'd be reasonable for me to just add that flag for the tools that make it easy (like JAX and I assume PyTorch and/or TensorFlow)

It is harmless, although I don't really understand if disabling that flag is sufficient for JAX. All of the input loading and NumPy scaffolding code uses double precision, and the JAX flag was to prevent it from turning it into single-precision along the way. *Without* that flag, it may do some strange hybrid, depending on where the conversion happens.

samestep:
@athas I have already done both of those things; you can check by running this command as described in `CONTRIBUTING.md`:

```sh
bun run --filter=@gradbench/website dev
```

Changes to <https://gradben.ch/> are not deployed from PRs, only from `main`.

athas (approved):

PR #562, opened 2025-05-06
Support a `--precision` CLI argument for tools

samestep:
Followup on https://github.com/gradbench/gradbench/pull/561#issuecomment-2856202123.


samestep:
Closing this PR since in the aforementioned discussion I think we decided that we're not going to be implementing this for all tools anyways.

PR #563, opened 2025-05-12
Add CITATION.cff

athas:
This add a file containing metadata for citing GradBench. It is relevant because I have some students who will need to cite GradBench properly soon-ish. The citation file only covers GradBench as a software artifact, not any papers that we will likely write based on it.

I simply added everyone who has commits in GradBench as an author, ordered by their first commit. This seems the most fair. It cannot be fully automated, because people will eventually make commits using different names, email addresses, or anonymously. However, I don't expect maintenance of this file to be particularly onerous. It is also not a permanent record, so we can add ORCIDs and such over time.

If this is accepted, a good next step is to also upload GradBench to [Zenodo](https://zenodo.org/). There are GitHub actions for automating this.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,6 +10,7 @@
   },
   "[typescript]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
   "[typescriptreact]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
+  "[yaml]": { "editor.defaultFormatter": "esbenp.prettier-vscode" },
   "editor.formatOnSave": true,
   "files.associations": {
     "time.h": "c"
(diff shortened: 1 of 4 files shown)


samestep (approved):
Awesome, thank you! This all seems very reasonable. I don't know anything about Zenodo, so I don't really have an opinion about adding a GitHub Action for that.

Could I request that you add a couple additional things in this PR?

1. Write a short section in `README.md` pointing to this file, similar to how many GitHub projects list BibTeX code in their README.
2. Write a short bit in `CONTRIBUTING.md` telling contributors that if they open a PR, they should add make sure to themselves to this file if they haven't already.

athas:
Oh, the web UI already has BibTex conversion. That's handy.

samestep:
Oh gotcha, thanks! I didn't realize GitHub automatically adds a sidebar entry for this; in that case you don't have to modify the `README.md` here unless you still want to.

athas:
I have done these things.

samestep:
Perfect, thanks!

Issue #554, opened 2025-04-25
Difficult to run all linters/formatters


samestep:
Currently we enforce the following linters/autoformatters on the codebase; in no particular order:

- [Prettier](https://prettier.io/)
- [clang-format](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormat.html)
- [Ruff](https://docs.astral.sh/ruff/)
- [rustfmt](https://github.com/rust-lang/rustfmt)
- [Clippy](https://doc.rust-lang.org/clippy/)
- [markdown-toc](https://github.com/jonschlinkert/markdown-toc)

We're also considering adding another Markdown linter in #333.

In CI, we spread this across [four jobs named `check-*`](https://github.com/gradbench/gradbench/blob/b241866b2c90802d45f7d67077ae7c9af6155455/.github/workflows/build.yml#L13-L66). There is no easy way to run all these tools locally.

It would be helpful to create a `repo lint` subcommand that runs all of these tools.

athas:
Both Futhark and Haskell also have opinionated formatters we could add. It'll be a right circus.

samestep:
Huh, `steam-run` is still necessary? That's weird... I thought #399 fixed all the Python issues we were having? I guess I must have forgotten if `ruff` still doesn't work properly?

PR #567, opened 2025-05-12
Enable markdownlint

### LINKED ISSUE #333: Markdown is not linted

samestep:
Resolves #333.


athas:
Worsens #554.

PR #568, opened 2025-05-12
Define the GMM objective function mathematically

samestep:
This PR attempts to improve `evals/gmm/README.md` by actually providing a mathematical definition of the function to be computed, fixing some errors from the original ADBench along the way. It also makes a couple small tweaks to the definition of the function input, for greater convenience.


athas:
This is lovely work. I hope the other ADBench-derived READMEs are not as faulty, but I have no expectation that they are not.

Do you want assistance updating the various implementations to behave as per the newly correct specification?

samestep:
Thanks @athas, I appreciate it!

And yes, it would be amazing if you'd be willing to help me update the implementations. Unfortunately we can't do that yet, because this is not _quite_ done; if you look in the implementation section, you'll see that I've reverted back the "simplification" of $`\text{tr}(\mathbf{\Sigma}_k)`$ which the ADBench authors had incorrectly converted to $`\|Q(\boldsymbol{q}_k, \mathbf{L}_k)\|_F^2`$. So now the issue is that we don't have a way to actually compute this trace.

I searched around and found various posts on Math Overflow and the Math Stack Exchange; it seems like various people have tried to do similar things, because this issue of parametrizing covariance matrices comes up a lot in computational statistics. So at least we're not doing something _too_ weird.

But, in any case, it's a bit unfortunate: as far as I can guess, the parametrization of $`\mathbf{\Sigma}^{-1}`$ via $\boldsymbol{q}$ and $\mathbf{L}$ and $Q$ is so that we can use $Q$ to easily compute the quadratic form needed in the multivariate Gaussian PDF. If that were the only thing we needed to do then we'd be golden; parametrizing only the inverse of the covariance matrix instead of the covariance matrix itself would be clearly the right choice.

However, since we also need to compute $`\text{tr}(\mathbf{\Sigma}_k)`$, it's now less clearly the right choice; as far as I can tell, we're going to end up doing something similar to inverting $Q$, or at least using $Q$ with a specialized triangular solver, to compute that trace. So at that point, why not just parametrize the covariance matrix itself, rather than its inverse?

I _think_ that even with this wrinkle, it's still the right choice to parametrize the inverse instead of the covariance matrix itself. This is because we use the multivariate Gaussian PDF a lot more (scales with both $N$ and $K$) than we use the Wishart PDF (just scales with $K$). So I think this parametrization is still fine, but I still need to write out an explicit way to compute that trace, since in any case it should not be necessary to actually compute and invert $`\mathbf{\Sigma}^{-1}`$.

I found a paper called ["Calculating selected elements of the inverse of symmetric positive definite matrices"](https://doi.org/10.1016/0167-9473(88)90092-8) that seems like exactly what we want here; it gives a nice clean algorithm in section 3. So then, my concrete next two steps would be:

1. Double-check that I'm not doing something wrong by making a lower triangular matrix $Q$ and then constructing $`\mathbf{\Sigma}^{-1} = Q^TQ`$ instead of $`\mathbf{\Sigma}^{-1} = QQ^T`$, or instead of making $Q$ upper-triangular instead; specifically I need to understand whether that is just a _convention_ of the [Cholesky decomposition](https://en.wikipedia.org/wiki/Cholesky_decomposition), or actually required to achieve the right properties.
2. Look to see if popular linear algebra libraries like PyTorch etc already provide implementations for that algorithm in the paper I linked, or generally see if it'd be viable to implement this trace computation efficiently in the tools we have in this repo.

Thoughts? Does this all seem reasonable? Any parts you'd be more or less interested in?

samestep:
Oh!! Actually, I just realized: there's an easier way. Equation 5 from [the ADBench paper](https://arxiv.org/pdf/1807.10129) pretty much stated that they were using the Wishart PDF on the covariance matrices. However, instead of doing that, it looks like their _actual_ implementation uses the Wishart PDF on the _inverses_ of the covariance matrices (I think the technical term is "precision matrix"). And actually, from running it locally just now, it _seems_ like

```math
\text{tr}(\mathbf{\Sigma}_k^{-1}) = \|Q(\boldsymbol{q}_k, \mathbf{L}_k)\|_F^2
```

is a totally fine simplification; I'll double-check the algebra in just a bit.

But if this works, then there's no need to do anything fancy here, I'll just modify the definition to say that the prior uses the Wishart distribution on the precision matrices instead of the covariance matrices.

I'll double-check this now and if it works then I'll update the rest of the math to reflect it.

samestep:
OK, I've done that; convenient that the math ended up working out.

Now once I flip the sign for the prior in all the existing implementations (very easy), all that'll be left to do is to modify the input/output for the existing implementations to match the new TypeScript I've written here. Hopefully that shouldn't be too bad.

I think before I do that, though, I'll undo my proposed change to have `l` be a ragged array. It's probably easier for it to just be how it was before, as flattened columns.

samestep:
Nope, I just wasn't sure how to do it. Thanks!

samestep:
OK @athas I've fully updated the PyTorch and JAX implementations; is there any chance you'd be able to help with some of the others? This is all the time I have for today, but I can also come back to this tomorrow.

Also, let me know if there's anything you're not a fan of in this PR as it currently stands.

athas:
I am somewhat uneased by the new requirement for sorting (the `argsort` stuff) in the objective function. That significantly changes the computational properties of this eval, as it is not about dense floating point arithmetic. Is there a good reason not to do this outside the objective function itself, similar to how the inverse covariance factors (`icf`) used to be computed directly by the eval?

samestep:
@athas to clarify, the only difference here is that I've split `icf` matrix has been split into its two conceptually distinct parts `q` and `l`. Previously `icf` was formed by concatenating `q` and `l` horizontally.

So, the same exact work needs to be done as before, with the only difference being that previously, one needed to ignore the first $`D`$ columns of `icf` when dealing with `l`, and now that is no longer necessary.

So yes, I'd be totally inclined to agree that my usage of `argsort` is probably not great; I wrote that code in a bit of a rush last night because I couldn't immediately figure out how to do it in the same way as the previous code, so I did this instead. But since `l` is literally just a subpart of what `icf` used to be, there should be no obstacle to doing computations with it in the same way as before, without `argsort`.

samestep:
So to put it another way: this actually _reduces the amount of work that needs to be done in the tool, because previously the eval would concatenate `q` and `l` together into `icf`, and then the tool would need to use slicing to pull them apart again (as you could see by the repeated use of slicing in the previous PyTorch and JAX implementations of this eval). My usage of `argsort` is just an artifact of me rushing last night.

samestep:
@athas I would not recommend constructing `icf`, as it's not a meaningful thing; it was just a hasty concatenation of `q` (which needs to be exponentiated and put in the diagonal) and `l` (which does not).

In the PyTorch code, `Ls` represents just the strictly lower triangular part of $`Q(\boldsymbol{q}_k, \boldsymbol{l}_k)`$ from the math; that is, if you just deal with $`\boldsymbol{l}_k`$ and leave out the diagonal that comes from $`\boldsymbol{q}_k`$. So, it really should be easier to get from `l` to `Ls` than it was to get from `icf` to `Ls`.

I'll now edit the PyTorch and JAX implementations to make them more similar to what they were before.

athas:
Alright, I think i have validation for the objective function - now for the task of rewriting the manually differentiated one.

athas:
I would prefer to have them in the TypeScript types, even if it is not syntactically valid. We do so in the documentation for other evals and it is much easier to understand that way. I personally find it very difficult to mentally track sizes any other way. I spent a year coming up with a type system for this stuff because I find it so annoying to keep track of.

samestep:
I don't think just putting the dimensions _directly_ into the TypeScript syntax would even be helpful. For instance, we have $`\mathbf{X} \in \mathbb{R}^{N \times D}`$; given that the brackets in TypeScript are postfix instead of prefix, I believe that would correspond to writing `x: Float[d][n]`, but then I'd guess it'd be very easy for people to misinterpret that as$`\mathbf{X} \in \mathbb{R}^{D \times N}`$.

samestep:
Honestly, if we're considering alternative syntax, I'd be more inclined to just use Futhark syntax directly. Would that be reasonable, do you think?

samestep:
Yes, I don't remember seeing any other examples in this repo using that modified TypeScript syntax for multidimensional non-square arrays (although I'm not 100% sure of that). But because of this potential for confusion, I do not want to put indices into the TypeScript syntax for GMM.

athas:
`manual` is now correct.

I can update many of the others. `ocaml` will be a challenge. I do not have the ability to fix `scilean`.

athas:
For Tapenade, I think constructing `icf` is the best choice. Modifying three different source files (two of which are generated) seems crazy.

samestep:
Sure, that seems reasonable given Tapenade's unwieldiness.

athas:
Is updating the tool implementations the only missing piece of this PR?

samestep:
@athas yes, I believe so.

athas:
Alright, I'll work on updating them. It will take a while however, as I have become busy with some other things, and untangling some of this code is unusually soul-crushing work.

samestep:
I'll handle the TensorFlow one.

athas:
Based on a quick skim, the Lean implementation seems to already implement the decomposition of `icf` into `q` and `l` (although with different names), so perhaps its modification is within my extremely meagre Lean skills.

athas (approved):
I'm hardly going to reject this after rewriting so much code.

samestep:
@athas I really appreciate all the work you put into this.

PR #569, opened 2025-05-13
lse.hpp: also parallelise second loop

athas:
No idea how I forgot to do that, especially since I even documented the resulting bad performance.


samestep (approved):
Cool.

PR #571, opened 2025-05-14
doc: add some more text for Julia and C++

athas:


cpp/README.md
```diff
@@ -52,3 +55,34 @@ equivalent of the `input` field of an `evaluate` message. This can be useful for
...
+to write the code such that it will run well when compiled for both sequential
+and multithreaded execution, ideally without implementing each function twice.
+
+OpenMP is fairly well suited for this purpose, as pragmas are (almost) ignored
+when `-fopenmp` is not passed to the C compiler, although we have encountered
+some wrinkles.
+
+### Reductions
+
+One problem is that Enzyme does not support the OpenMP reduction clause. This
```
wsmoses, line 71: oh I didn't know this wasn't supported, open an issue? It was supported for some version of clang/libopenmp

athas: There is/was already an issue: https://github.com/EnzymeAD/Enzyme/issues/1246

wsmoses: Yeah so that was closed so we added support, so it not working is news to us

athas: I think that issue was closed because it was migrated to a discussion: https://github.com/EnzymeAD/Enzyme/discussions/1251

If I write a program with a straightforward +-reduction I get the same error as in that issue. I can of course still open a new issue; let me just put together a self-contained program.

PR #573, opened 2025-05-14
ci: migrate linting into scripts

athas:
This makes it easier to run the various linters outside of CI. In particular, I have added a pre-commit hook that does so.

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -1,0 +1,24 @@
+#!/bin/sh
+#
+# Pre-commit hook for quality-checking the commit. Currently checks all files,
+# not just the ones touched by the commit. You can also run this script
+# directly.
+#
+# Install this hook by:
+#
+# $ ln -s ../../scripts/hooks/pre-commit .git/hooks/pre-commit
+
+fail() {
+    echo "Aborting commit due to verification errors."
+    echo "If you disagree, use git commit --no-verify."
+    exit 1
+}
+
+echo Quality-checking commit...
+echo
+
+for l in scripts/lint/*.sh; do
+  echo $l
+  sh $l || exit 1
+done
+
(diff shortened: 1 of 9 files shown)


scripts/hooks/pre-commit
samestep, line 1: This seems fine to have for people who like pre-commit hooks (personally I don't like them, but nothing wrong with others using them). I'd add a mention to it in `CONTRIBUTING.md` though, since otherwise it's not very discoverable.

scripts/lint/README.md
samestep, line 1: Again, I'd really rather these scripts be incorporated into the CLI so they're not a separate special thing; and in either case, it should be documented in `CONTRIBUTING.md`.

samestep (commented):
I'd personally really rather incorporate this into the CLI via a `repo lint` subcommand as I suggested in #554; I don't have a problem with doing that myself, but I need to sit down and actually do it.

athas:
I have no issue with you putting these into `gradbench` itself, although I'm skeptical of the advantage. It is just more verbose. It may also make ad-hoc control flow more difficult (e.g. I'm considering adding some NixOS-specific workarounds) to the commands that run `ruff`.

samestep:
In that case, should we be using Rust for the CLI at all? If it's just more verbose, would it be better to just use shell scripts for the whole thing?

samestep:
(That's what I was doing originally; maybe that was the right approach and I got overzealous?)

athas:
Different tools for different jobs. Some things call for shell scripts.

scripts/hooks/pre-commit
athas: Done.

scripts/lint/README.md
athas: This is now done. I really don't mind this eventually being integrated into `gradbench` if you think there is benefit to it, but this collection of scripts addresses an immediate problem (that it's hard to run all the linters), and doesn't prevent us from doing better in the future.

samestep:
Almost done with #574, which should supercede this.

PR #574, opened 2025-05-14
Add a `repo lint` CLI subcommand

### LINKED ISSUE #554: Difficult to run all linters/formatters

### LINKED ISSUE #576: Add details to `CONTRIBUTING.md` on using Prettier
I struggled for a bit with `Prettier` installation and use for #575. If others are in my situation, it might be useful to know that the following sequence of things need to be done:

1. Install a Node version manager like `nvm`: https://github.com/nvm-sh/nvm
2. Install `Node.js` and `npm`: https://docs.npmjs.com/downloading-and-installing-node-js-and-npm
3. Install `Prettier`: https://prettier.io/docs/install
4. Install `prettier-plugin-organize-imports`: https://www.npmjs.com/package/prettier-plugin-organize-imports

samestep:
Resolves #554 and hopefully also resolves #576.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,63 +10,35 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
-  check-cpp:
+  cli:
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Install clang-format 19
-        run: sudo apt-get install -y clang-format-19
-      - name: Check C/C++ formatting
-        run: |
-          git ls-files '*.c' '*.cpp' '*.h' '*.hpp' | xargs clang-format-19 --dry-run -Werror
+      - name: Build CLI
+        uses: houseabsolute/actions-rust-cross@v1
+        with:
+          target: x86_64-unknown-linux-musl
+          args: --package=gradbench --profile=release-with-debug
+      - name: Upload CLI as artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: cli
+          path: target/x86_64-unknown-linux-musl/release-with-debug/gradbench
 
-  check-js:
+  lint:
+    needs: cli
     runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Install Bun
-        uses: oven-sh/setup-bun@v2
-      - name: Install npm dependencies
-        run: bun install
-      - name: Check Markdown tables of contents
-        run: |
-          bun run toc
-          CHANGES=$(git status --porcelain)
-          echo "$CHANGES"
-          git diff
-          [ -z "$CHANGES" ]
-      - name: Check Prettier formatting
-        run: bun run prettier --check .
-      - name: Run TypeScript
-        run: bun run --filter='*' typecheck
-
-  check-py:
-    runs-on: ubuntu-24.04
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-      - name: Setup Python
-        uses: actions/setup-python@v5
-      - name: Install uv
-        uses: astral-sh/setup-uv@v5
-      - name: Lint Python code
-        run: uv run ruff check
-      - name: Check Python formatting
-        run: uv run ruff format --check
-
-  check-rs:
-    runs-on: ubuntu-24.04
-    steps:
-      - name: Checkout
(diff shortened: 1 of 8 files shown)


gdalle:
If you want to add Julia files, the reference formatter is https://github.com/domluna/JuliaFormatter.jl

The corresponding command would be

```bash
julia -e 'using Pkg; Pkg.add("JuliaFormatter"); using JuliaFormatter; format(".")'
```

We can get more sophisticated by specifying the version of everything or adding formatting options or caching the installation, but that's a good start

samestep:
@gdalle awesome, thanks! Let's do that in a followup PR, because I suspect that it'll involve making edits to various Julia files already in the repo, and I want to keep this PR scoped to just consolidating the lints we already have.

samestep:
OK, I think this is ready for review.

@gdalle, do the revisions I've made here to `CONTRIBUTING.md`, along with the links and suggested commands in the error messages printed by `repo lint`, resolve the discoverability issue you raised in #576?

@athas, judging from #573 it seems like `uv run ruff` still doesn't work for you; what would be your preferred way for me to handle that in this PR?

gdalle:
looks good to me, thanks!

athas (approved):

samestep:
@athas I've pushed 45c0ec4f267216ee0574c6cec7d38015d0039298; does this now allow you to just run `./gradbench repo lint` without having to deal with `TMPDIR` or `steam-run`?

athas:
No, I would still need to set (or unset) `TMPDIR`.

samestep:
@athas why is that? I thought you said that it's only about the interaction between `TMPDIR` and `steam-run`, but the lints that use `steam-run` are disjoint from the lint that uses temporary files. Or am I wrong?

Could you print the error you get when running it currently without setting `TMPDIR`?

samestep:
To clarify: I tried it on my own NixOS machine now, and I actually _do_ get errors when running it on, e.g., commit 687189c88c4b718965a5cd4ae0b0af100a2bb95c (that is, without `steam-run`), so it seems that the mysterious difference between our machines has disappeared. But I get no errors running `./gradbench repo lint` on my NixOS machine on commit 4aeb167807e8c1b79440fc68caaa82d3e763fe73, for instance.

athas:
I believe `ruff` uses temporary files internally, and uses `TMPDIR` for that purpose, and insists that that directory exists in its entirety. The pertinent part of the error is this:

```
running Ruff linter
error: No such file or directory (os error 2) at path "/tmp/nix-shell-424456-0/.tmpqvy5Ng"
 failed Ruff linter

running Ruff formatter
error: No such file or directory (os error 2) at path "/tmp/nix-shell-424456-0/.tmplgIYov"
 failed Ruff formatter
```

Note here that `TMPDIR=/tmp/nix-shell-424456-0`, which is a directory that does not exist inside `steam-run` (but `/tmp` *does* exist).

samestep:
Bizarre. OK, I guess the mysterious differences between our NixOS machines remain, haha.

How about with 771a08de4b82598c1a1e0f6123bdf4bef9b6bf1d?

samestep:
Sure, I'd be happy to go open a bug on the Ruff repository if no other NixOS user has already opened one; can't remember if I've looked already.

So for this PR, what specifically would you prefer for us to do?

samestep:
For instance, for the `steam-run` issue, there's astral-sh/ruff#1699, but it seems like that's not really considered to be a bug on Astral's side.

athas:
Merge it. It works well enough for me, and I'm not averse to coming up with my own workarounds. The nice thing about having things that fail on NixOS is that NixOS users are used to fixing their own problems.

PR #575, opened 2025-05-15
Add ForwardDiff, ReverseDiff and Mooncake through DI (take 2)

gdalle:
Supersedes #498, replaces #479 

This PR adds 3 new tools:

- ForwardDiff.jl
- ReverseDiff.jl
- Mooncake.jl

for 3 evals:

- hello
- det
- ode


vchuravy (approved):
Looks very nice!

gdalle:
We have a problem here:

```
.github/workflows/build.yml (Line: 208, Col: 7): Strategy for job 'run' produced 276 configurations which exceeds the maximum of 256 configurations
```

athas:
How is CI passing? What happened to the matrix issue?

athas:
Ah, that's annoying. We unfortunately cannot merge this until we come up with a solution for the matrix issue, as it would prevent the nightly builds from working.

athas:
For the plots to work you will also need to add some colors to this table: https://github.com/gradbench/gradbench/blob/7f1e6ea6e10f590dcc3370ae71ee1df3c54db69b/js/website/src/Stats.tsx#L9

(And they still won't actually work until after the PR is merged, as the website runs off `main`.)

athas (approved):

athas:
I will merge this if CI passes.

PR #578, opened 2025-05-15
Clarify formatting in CONTRIBUTING.md

gdalle:
I'm not sure if the previous "JavaScript" section also applied to actual tool code written in JavaScript? If so, we might need yet another section title

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,24 +2,25 @@
 
 <!-- toc -->
 
-- [Setup](#setup)
-- [Dependencies](#dependencies)
-- [CLI](#cli)
-- [Docker](#docker)
-  - [Multi-platform images](#multi-platform-images)
-- [Tools](#tools)
-  - [Implementing a new eval for a tool](#implementing-a-new-eval-for-a-tool)
-- [Evals](#evals)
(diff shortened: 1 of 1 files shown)


samestep:
> I'm not sure if the previous "JavaScript" section also applied to actual tool code written in JavaScript? If so, we might need yet another section title

It does, yeah. Currently we have two tools with code written in JavaScript (TypeScript, to be precise).

Here is a list of all the languages Prettier formats: https://prettier.io/docs/

So yeah, I'm not exactly sure the best way to organize this; currently I just have it organized by what _ecosystem_ all the tools live in, and Prettier just happens to exist in the JavaScript ecosystem. But as you've mentioned, by touching Markdown, it kind of extends "beyond its ecosystem".

Thoughts?

gdalle:
I think just adding this note to the top is enough

samestep:
I'm almost done with #574, so let's hold off on this for a bit; we'll want to update `CONTRIBUTING.md` once we have the `repo lint` subcommand, so I'm thinking it might make sense to put that in its own section. And then as for installing all the individual linters, I'm thinking the discoverability can be improved by just having `repo lint` print installation instructions for any tools that it doesn't find.

PR #583, opened 2025-05-16
Allow multiple eval/tool pairs in a single CI job

### LINKED ISSUE #365: Nightly doesn't check intermediary exit code
In #247 we added some Bash code to our non-nightly build so it checks that the exit code from `gradbench run` is what we expect, for each eval/tool pair. However, I didn't do this for nightly, for a couple reasons:

1. It was just gross shell code, and I didn't want to duplicate it.
2. I kind of wanted the nightly build to still succeed regardless of the outcome for each eval/tool pair.

As of #342 and #290, this isn't great, because now the `gradbench stats` command (and thus also website) use the _expected_ outcomes in their logic for determining colors of squares in the grid, instead of the _actual_ outcomes. This is fine for our non-nightly builds because we assert those are the same, but it could potentially cause some weird behavior for nightly builds in the future if they drift out of sync. We should _either_

1. assert that the expected and actual outcomes match in the nightly build too, _or_
2. modify CI so that the grid chart logic only uses the _actual_ outcomes, not the expected ones.

Probably the second option is better, because this would also move us more toward charting infrastructure that can be used outside of the specific set of eval/tool pairs currently in the GradBench repo, relating to #363.

### LINKED ISSUE #577: GitHub Actions matrix is limited to 256 items
This is a problem for #575 which finally brings the product of evals and tools above that limit.

Here are two ways to limit the number of items and why they won't work well:

1. Use one item per eval. Since we don't have that many evals, this will make each item take a long time to run.
2. Use one item per tool. This will result in very unbalanced load, as the slowest tools are much slower than the fastest tools. Also does not exploit concurrency well.

One solution is semi-curated clustering, where tools that are known to be fast are bunched together in fewer items.

Another is to cheat and try to fake bigger matrices using things like [Matrix extended](https://github.com/marketplace/actions/matrix-extended), but I have heard stories of [this being treated as abuse](https://github.com/orgs/community/discussions/38704), so I am a bit wary of that approach.

Since the logic for generating the matrix is done in Rust in the `gradbench` CLI, I think we can get away with putting fairly intelligent (or ad hoc) rules there, so that is what I prefer.

samestep:
Resolves #365, resolves #577.

This PR adds flags `--download-github` and `--check` to the `repo run` subcommand so that we can use it in CI to run multiple pairs in a single job, and modifies the `matrix` and `stats` jobs to accommodate this change.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -186,37 +186,18 @@ jobs:
         uses: ./.github/actions/space
       - name: Install CLI from artifact
         uses: ./.github/actions/cli
-      - name: Download eval Docker image from artifact
-        uses: actions/download-artifact@v4
-        with:
-          name: eval-${{ matrix.eval }}
-      - name: Download tool Docker image from artifact
-        uses: actions/download-artifact@v4
-        with:
-          name: x86-tool-${{ matrix.tool }}
-      - name: Load eval Docker image
-        run: docker load --input eval-${{ matrix.eval }}.tar
-      - name: Load tool Docker image
-        run: docker load --input tool-${{ matrix.tool }}.tar
-      - name: Run tool on eval
+      - name: Run
         id: run
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          set +e
-          gradbench exit-code ${{ matrix.outcome }}
-          expected=$?
-          gradbench run --eval "gradbench eval ${{ matrix.eval }}" --tool "gradbench tool ${{ matrix.tool }}" --output log.jsonl --timeout 600
-          actual=$?
-          set -e
-          if [ $actual -ne $expected ]; then
-            echo "expected exit code $expected (from outcome ${{ matrix.outcome }}) but got exit code $actual"
-            exit 1
-          fi
-      - name: Upload log as artifact
+          gradbench repo run --download-github ${{ github.run_id }} -o run --timeout 600 ${{ matrix.args }}
+      - name: Upload logs as artifact
         if: success() || steps.run.conclusion == 'failure'
         uses: actions/upload-artifact@v4
         with:
-          name: run-${{ matrix.eval }}-${{ matrix.tool }}
-          path: log.jsonl
+          name: run-${{ matrix.artifact }}
+          path: run
 
   stats:
     needs:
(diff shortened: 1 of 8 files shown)


gdalle:
Once this backend infrastructure is in place, where are the groupings actually decided for CI?

samestep:
@gdalle my current plan is to just keep pretty much the same `matrix` CI job that we already have:

https://github.com/gradbench/gradbench/blob/cc3008560dfe3b7df8c5b581d5e13a9abd017072/.github/workflows/build.yml#L100-L114

The implementation for that `repo matrix` CLI subcommand is here:

https://github.com/gradbench/gradbench/blob/cc3008560dfe3b7df8c5b581d5e13a9abd017072/crates/gradbench/src/main.rs#L1083-L1119

So, I intend to just change it so instead of printing a list of eval/tool pairs, it will print a list of strings that will then just be used as CLI args for the `repo run` subcommand.

It seems like the simplest thing to do would be to just have 256 "slots" which get filled up with eval/tool pairs, and then once all the slots are filled, go back to the beginning and add more pairs to existing slots.

But that probably won't be _quite_ good enough, because as you may have seen, we currently have a CI step to delete some stuff to free up disk space:

https://github.com/gradbench/gradbench/blob/cc3008560dfe3b7df8c5b581d5e13a9abd017072/.github/workflows/build.yml#L185-L186

So if that naive process results in a slot having, e.g., two pairs that have different evals _and_ different tools, I suspect that'd use too much disk space. So maybe it'd need to do something like, only add a pair to a slot that shares the same tool (because Docker images for tools tend to be bigger than those for evals).

athas:
> So if that naive process results in a slot having, e.g., two pairs that have different evals and different tools, I suspect that'd use too much disk space.

I have a hunch that might not be as big a problem as one might fear - our Docker images are not actually very large, and we haven't even put any real effort into shrinking them. The stuff that they *must* contain is actually quite small, if we remove unnecessary build products. We (or I) just have not bothered to do so yet.

I recall that the disk space problems were mostly due to the log files (which *will* also increase in aggregate size once we add fourteen Julia tools), not the Docker images.

samestep:
@athas I think you're largely correct, but specifically the SciLean image is massive.

But generally I agree that it's not going to be a big deal either way. The thing I proposed about grouping together pairs that share a tool is not going to be hard to implement.

gdalle:
For what it's worth, we're not gonna add 14 Julia tools here. I think that (Polyester)ForwardDiff, Enzyme, ReverseDiff, Zygote and Mooncake are a sufficient subset. The rest is either finite differences, symbolic approaches, or too old to matter.

gdalle:
I don't write Rust so my review probably wouldn't be useful here, but I have merged this branch into #575 to see if it solves the issue

Issue #406, opened 2025-03-27
No plots for particle and saddle on the website


athas:
These need special care as they cannot be plotted like our our benchmarks.

samestep:
Right, as I mentioned in https://github.com/gradbench/gradbench/pull/378#pullrequestreview-2694990140:

> we already keep `Scorer` implementations separate per eval; the fact that we don't already do this in the website too is just because I didn't get around to implementing it yet. Going forward we really should have the plotting code dispatch based on the eval; this also relates to https://github.com/gradbench/gradbench/issues/363.

athas:
I'm working on the scoring, but I will need help with the plotting.

PR #638, opened 2025-10-25
Add `horde-ad` tool

Mikolaj:
The companion ticket in the horde-ad repo is at https://github.com/Mikolaj/horde-ad/issues/115.


samestep:
Would it be helpful if we perhaps maintained a Markdown table near the top of `README.md` with two columns like this?

1. the name of the tool in GradBench, which would also be formatted as a hyperlink to its own `README.md` under the `tools` directory
2. a one-line description of the tool, which could include the programming language or anything else deemed the most important

That way it'd be easy for anyone to see that this is a Haskell tool even if they don't click any links.

athas:
I am also OK with renaming `haskell` to `haskell-ad`. I don't remember why I called it just `haskell`.

Mikolaj:
> I think that is a stale comment. At one point I tried all kinds of tricks to fail in a more graceful manner when `ad` would run out of memory.

OK, thank you. Shall I fix this comment for the haskell tool in this PR? (I already make a couple of minor tweaks, but I can separate them out, if you prefer.)

samestep:
> @samestep: shall I rename haskell to haskell-ad in this PR?

Sure, go for it!

Mikolaj:
Done and done (and done).

js/website/src/components/EvalViz.tsx
```diff
@@ -17,7 +17,8 @@ const colors = {
   floretta: "#f10537",
   "forwarddiff-jl": "#4063d8",
   futhark: "#5f021f",
-  haskell: "#5e5086",
+  haskell-ad: "#5e5086",
+  horde-ad: "#5e5086",
```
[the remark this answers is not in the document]
samestep: Also, could you ensure they have different colors?

Mikolaj:
> If you want `det` to be less slow, you could instead implement the algorithm used by the Futhark implementation: https://github.com/gradbench/gradbench/blob/main/tools/futhark/det.fut

Interesting. i will try.

> It is fundamentally less efficient than the recursive one (yes, it is possible), but it does not rely on constructing quite so many small structures.

Even less efficient? Amazing! Yes, I guess (though I haven't profiled) the allocations of all the tensor meta-data (and GC?) is what was driving the cost above that of haskell-ad when I was using exactly the same code as haskell-ad but with the element type set to rank 0 tensors. In the new faster symbolic version it's probably the tensor meta-data plus the build-up of AST terms from the unrolled recursion plus whatever overhead comes from handling `rbuild1` (maybe none? these are just mapping in this case).

athas:
Which draft is that? The gamma functions are from the C math library, and just defined like this: https://github.com/diku-dk/futhark/blob/ba5b1ed4f81e8966b633aeeda329f75a401a2a63/src/Futhark/Util/CMath.hs#L84-L106

You can extract testing data from the log files produced by the runner. I don't think I have any other validation data.

Mikolaj:
Thank you for the tips. I meant this draft: https://github.com/gradbench/gradbench/blob/main/tools/haskell/src/GradBench/GMM.hs and, in particular, if you maybe tested it the way I do it or similar (`PYTHONPATH=python/gradbench/:$PYTHONPATH ./gradbench repo run --eval "$ uv run python/gradbench/gradbench/evals/saddle/run.py" --tool "$ cabal run --project-dir tools/horde-ad gradbench"`), which requires the ability to convert the bunch of C arrays to Haskell (it's probably easy to code, but I'm eager to copy-paste instead whenever I can :).

Mikolaj:
That's very helpful, thank you. Who would have thought there is an actual documentation?

Mikolaj:
@athas: in the end, your Futhark implementation was the most helpful of all, but all the other sources of information helped. I was wondering about GMM for haskell-ad, but it would be terribly slow, wouldn't it? I don't know `ad` enough to be sure. From my code at least the JSON instances could be reused. ;)

Mikolaj:
I still need some GHC changes to profile and optimize the tool properly, but I hope this should at least work and produce some results.

athas:
Great to see this finally running! A fair reward after having to struggle with an automated Markdown formatting pedant.

athas:
I suspect `kmeans` will also time out.

Mikolaj:
Yes, I'm very glad it even runs. Thank you for your help and for so much code to copy and tweak. :)

Regarding `kmeans`, yes, it will timeout, it did previously, too, I only renamed the haskell-ad tool, not tweaked it (except for a minor detail somewhere), so it's not fixed.

athas:
Oh right, I mixed it up with `horde-ad`, for which `kmeans` runs perfectly well.

Mikolaj:
I think this is ready for review now.

tools/futhark/lstm.fut
```diff
@@ -20,17 +20,17 @@ def lstmModel [d]
   let hidden2 = map2 (*) outgate <| map f64.tanh cell2
   in (hidden2, cell2)
 
-def lstmPredict [slen] [d]
-                (mainParams: [slen][2][4][d]f64)
+def lstmPredict [stlen] [d]
```
athas: While this does improve consistency in the Futhark code, I'd rather prefer if this PR was concerned solely with Haskell.

tools/horde-ad/cabal.project
```diff
@@ -0,0 +1,4 @@
+packages: gradbench.cabal
```
athas, line 1: I would prefer using `index-state` to ensure reproducibility.

athas (changes requested):
I do not have the expertise to judge whether the use of `horde-ad` is good (but who else but the author would?), and I am not going to quibble about specifics of Haskell style. Apart from two specific remarks, I have the following considerations, none of which should be considered requests:

* Would it be worthwhile to share more of the driver code between the two Haskell tools? The reason I don't think it is *certainly* worth it is because it is unlikely we will have more than two Haskell-based tools, and sometimes the conceptual and technical overhead of deduplication exceeds the gain of shared code - especially since GradBench does not have any general mechanism for sharing code between tools.
* I would appreciate a comment in each implementation reflecting on how good it is, and any subtle reasons `horde-ad` does well or badly on a particular eval. This is extremely useful when the benchmark suite is used by others, and they have to explain results in the future. More general notes on the suitability of `horde-ad` for GradBench's benchmark suite would also be appreciated (in the tool README).

tools/futhark/lstm.fut
Mikolaj: Commit removed.

tools/horde-ad/cabal.project
Mikolaj: Makes sense. Done.

Mikolaj:
Regarding the sharing of the Haskell driver code, the only file with non-trivial size is Main.hs at 200 lines, of which 8 lines differ (I've just pushed a commit that minimizes the diff). IMHO it's easier to maintain 2 nearly identical files than however clever system of sharing fragments of code.

I will add comments about horde-ad vs particular benchmarks. Is there an easy way to see the benchmark results with horde-ad included? If not, maybe I'd add these comments in a followup PR?

athas:
Unfortunately, the automation doesn't work for PRs coming from non-committers, so we won't see the full results before merging.

athas (approved):

Mikolaj:
In that case, let me make a followup PR once I see the exciting results. :)

Thank you for the review!

Mikolaj:
When `Build / tool (horde-ad, true) (pull_request)` is run, I'm seeing both
 ```
[linux/arm64 5/5] RUN cabal build --allow-newer --project-dir tools/horde-ad gradbench
```
and
```
[linux/amd64 5/5] RUN cabal build --allow-newer --project-dir tools/horde-ad gradbench
```
Why are both needed?    

Regardless, something is stuck this time and it takes 2h already.

**Comment by athas** [comment]
Why both arm64 and amd64? I don't remember; I think Sam just wants to support multiple architectures.

File: tools/horde-ad/src/GradBench/BA.hs [hunk 4]
```diff
@@ -0,0 +1,52 @@
+-- TODO
+-- See https://github.com/tomsmeding/ADBench/blob/accelerate/tools/Accelerate/src/BA.hs and https://tomsmeding.com/f/master.pdf and https://github.com/gradbench/gradbench/tree/main/evals/ba
```
samestep: Would it be OK to modify this comment slightly? I'm a bit allergic to GitHub links that point to branches instead of commits.

```suggestion
-- See evals/ba in this repo, as well as https://github.com/tomsmeding/ADBench/blob/101508da93cd4fc80c0ea7a690c6de824c815ca6/tools/Accelerate/src/BA.hs and https://tomsmeding.com/f/master.pdf
```

tools/horde-ad/gradbench.cabal
```diff
@@ -0,0 +1,86 @@
...
+name:               gradbench
+-- The package version.
+-- See the Haskell package versioning policy (PVP) for standards
+-- guiding when and how versions should be incremented.
+-- https://pvp.haskell.org
+-- PVP summary:     +-+------- breaking API changes
+--                  | | +----- non-breaking API additions
+--                  | | | +--- code changes with no API change
+version:            0.1.0.0
+synopsis:           Gradbench entry to horde-ad. A lot of this code is copied from https://github.com/gradbench/gradbench/tree/main/tools/haskell-ad and https://github.com/tomsmeding/ADBench/blob/accelerate/tools/Accelerate/src/
```
samestep:
```suggestion
synopsis:           GradBench entry to horde-ad. A lot of this code is copied from tools/haskell-ad in GradBench, as well as https://github.com/tomsmeding/ADBench/tree/101508da93cd4fc80c0ea7a690c6de824c815ca6/tools/Accelerate/src
```

samestep (approved):
This is great! Thanks @Mikolaj for putting all this work in :)

I'm happy to merge this as long as CI passes. Ideally the arm64 build works for horde-ad; if not then as a last resort we could add it as a second exception on this line:

https://github.com/gradbench/gradbench/blob/f26651a7bcc120654fa895adecaf371b8cc059c5/crates/gradbench/src/main.rs#L1266

Other than that, I just have a couple minor nitpicks about URLs in comments.

Mikolaj:
@samestep: both suggestions gratefully accepted. :)

Mikolaj:
Hmm, strange that it failed. It succeeded yesterday. That's the test where haskell-ad goes over the RAM limit, I think.

athas:
Let's see if we can get it to work, but otherwise I'll merge anyway, since it has nothing to do with this PR.

Mikolaj:
It seems the compilation of horde-ad on arm64 works fine, but sometimes takes hours (4h this time). One potential reason is low (and maybe different in different runs, which would explain why it sometimes works much faster if I'm not mistaken) amount of RAM available, which may lead to a use of swap space, etc.

Mikolaj:
Given that, @samestep: shall I add horde-ad as the exception or is there a way to increase RAM in the arm64 workers (amd64 seem fine) and check this fixes the problem? Or maybe RAM available is the same, but binaries or binary representations of some things are larger on arm64 than amd64?

Mikolaj:
(Here's the job in question: https://github.com/gradbench/gradbench/actions/runs/24554399405/job/71808297468?pr=638)

athas:
I'm OK with merging this as-is. I vaguely remember that the ARM builds sometimes/always ran simulated with qemu, which is of course dreadful.

samestep:
Yep I'm just gonna merge this. :) Perhaps afterward I'll investigate ways to make the horde-ad QEMU build faster.

Mikolaj:
Splendid. Let me know whenever I can help.

Mikolaj:
BTW, this job failed again, though I'm not sure if that's what blocks the display of the new results: https://github.com/gradbench/gradbench/actions/runs/24572922151/job/71855567119

samestep:
Yeah that'd block it. I've now queued that job to re-run.

Mikolaj:
> Let's see if we can get it to work

@athas: I tried my best, but only got a few percent speed-up: https://github.com/gradbench/gradbench/pull/665

athas:
We finally managed to get a successful run, and the results are now on https://gradben.ch/

I don't know why it was so flaky for the past couple of days. The logs suggest that the artifacts don't work right.

Mikolaj:
Thank you very much. The suspense was killing me. :D 

Darn, I don't beat futhark even on a single test. I will have to try harder.

BTW, I probably overlooked this in the documentation --- what is the time shown in the cells of the table? Is it the sum over all data sizes of a single runtime of both primal and derivative run? Or not a single runtime, but sum or average of several runs?

BTW2, the graphs on https://gradben.ch/ are very informative, great job! But is it known that the graphs for particle and saddle are missing (I tried on both Firefox and Chromium)?

athas:
I also forgot what the cell timings are. I think it's total runtime, but I'm not sure.

The missing plots are [known](https://github.com/gradbench/gradbench/issues/406). It's probably not very difficult work, but somewhat tedious. Maybe an LLM can do it...

PR #644, opened 2026-01-16
Migrate from Docker to Nix

samestep:
Vibe coded, just to get a sense of how feasible this is. Haven't reviewed yet.


Mikolaj:
Yay, horde-ad/lse again has the speed it should have (meaning, on par with C implementations) and not a twice slower speed (it only had the good speed once in dozens of the old tests).

samestep:
Huh! That is interesting, I wonder why. I'd be kinda surprised if this is actually due to the difference in how we build things in this branch, but I haven't actually reviewed the code yet so I suppose it's possible. Do you have any hypotheses?

(In case it's useful, I asked Claude about this, and Claude's hypothesis is that the Docker build uses `-mtune=native` which could cause performance degradation if the build job has a different CPU from the run job. Sounds somewhat plausible to me, but still fairly surprising.)

Mikolaj:
> Huh! That is interesting, I wonder why. I'd be kinda surprised if this is actually due to the difference in how we build things in this branch, but I haven't actually reviewed the code yet so I suppose it's possible. Do you have any hypotheses?

I know too little about your system to hypothesize. I know that, in a local run, the results were skewed by julia jobs hogging cores and RAM, but that probably doesn't apply here, because you are using different github machines for different jobs?

> (In case it's useful, I asked Claude about this, and Claude's hypothesis is that the Docker build uses `-mtune=native` which could cause performance degradation if the build job has a different CPU from the run job. Sounds somewhat plausible to me, but still fairly surprising.)

I doubt this explanation. I specifically force  `-mtune=native` [here](https://github.com/gradbench/gradbench/blob/418efa165f690ae8d5f3684398afdb424e918171/tools/horde-ad/cabal.project#L19) in the desperate hope that once in a while the CPUs will align and I will see the 30ms runtime again, so I don't think I depend on any default that Docker or anything else sets up. I'm also not sure how much  `-mtune=native` changes, because I don't see big differences locally (but I have an 8-core AMD Ryzen, so probably very different from what github uses).

PR #663, opened 2026-03-12
New tool: autodiff.

athas:


athas:
The `llsq` implementation is so slow that I am concerned I am not using the tool right.

samestep:
Codex recommended this patch which seems to make it about 20% faster:

```diff
diff --git a/tools/autodiff/llsq.cpp b/tools/autodiff/llsq.cpp
index 8d343ae0..4eff492e 100644
--- a/tools/autodiff/llsq.cpp
+++ b/tools/autodiff/llsq.cpp
@@ -8,25 +8,30 @@ using namespace autodiff;
 
 class Gradient : public Function<llsq::Input, llsq::GradientOutput> {
 public:
-  Gradient(llsq::Input& input) : Function(input) {}
+  Gradient(llsq::Input& input) : Function(input) {
+    size_t n = _input.n;
+    size_t m = _input.x.size();
+    x_.resize(m);
+    for (size_t i = 0; i < m; i++) {
+      x_[i] = _input.x[i];
+    }
+    llsq::primal<var>(n, m, x_.data(), &y_);
+  }
 
   void compute(llsq::GradientOutput& output) {
-    size_t n = _input.n;
     size_t m = _input.x.size();
     output.resize(m);
 
     using Eigen::VectorXd;
-    VectorXvar x(m);
-    for (size_t i = 0; i < m; i++) {
-      x[i] = _input.x[i];
-    }
-    var y;
-    llsq::primal<var>(n, m, x.data(), &y);
-    VectorXd dydx = gradient(y, x);
+    VectorXd dydx = gradient(y_, x_);
     for (size_t i = 0; i < m; i++) {
       output[i] = dydx[i];
     }
   }
+
+private:
+  VectorXvar x_;
+  var        y_;
 };
 
 int main(int argc, char* argv[]) {
```

But obviously that doesn't close much of the gap at all. As far as I can tell, **autodiff** is simply in the category of tools that use bulky dynamic data structures instead of compilation and thus are just very slow.

samestep:
Also, hey, still twice as fast as Zygote! 🤣

athas:
I had already tried the trick of precomputing the augmented numbers, but the impact was small, and for most tools we don't do it that way. We're not completely consistent though, so maybe I should do it here just to give autodiff more of a chance.

athas:
The Codex proposal goes further than I am comfortable with, since it means the input isn't used at all in the `compute` method.

athas:
I think the core problem is that although this tool does support reverse mode, it's not the focus, and it will not handle problems of GradBench's scope well. The larger workloads for `lse` fail with a stack overflow, for this reason: https://github.com/autodiff/autodiff/issues/314

That does put somewhat of a damper on implementing most of GradBench, but I suppose I can look at the few evals that use only forward mode.

samestep:
> The Codex proposal goes further than I am comfortable with, since it means the input isn't used at all in the `compute` method.

Agreed.

> I think the core problem is that although this tool does support reverse mode, it's not the focus, and it will not handle problems of GradBench's scope well. The larger workloads for `lse` fail with a stack overflow, for this reason: [autodiff/autodiff#314](https://github.com/autodiff/autodiff/issues/314)
> 
> That does put somewhat of a damper on implementing most of GradBench, but I suppose I can look at the few evals that use only forward mode.

That makes sense, thanks for the extra context!

PR #665, opened 2026-04-17
Micro-optimize kmeans for haskell-ad slightly

Mikolaj:
I'm afraid I can't do better. In particular, I don't know if there's any better tool to use in this context than `hessianProduct'` nor how one can checkpoint in `ad`.

### DIFF
--- a/tools/haskell-ad/src/GradBench/KMeans.hs
+++ b/tools/haskell-ad/src/GradBench/KMeans.hs
@@ -7,24 +7,24 @@ module GradBench.KMeans
   )
 where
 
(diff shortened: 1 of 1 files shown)


Mikolaj:
@athas: if we currently have no ideas how to speed up kmeans for haskell-ad, then maybe let's add `+RTS -M12G` to the commandline so that it terminates after 12G (or whatever is the maximum we can handle) of heap is not enough? It could be as well added to horde-ad in case data sizes grow in the future and one of the benchmarks starts causing OOMs.

PR #667, opened 2026-04-24
Improve comments about horde-ad implementations

Mikolaj:
I will try understand the performance differences between horde-ad and others and then fix any trivial reasons of horde-ad slowness and comment about the rest. Along the way I will also generally improve my comments about horde-ad code.

### DIFF
--- a/tools/horde-ad/src/GradBench/BA.hs
+++ b/tools/horde-ad/src/GradBench/BA.hs
@@ -1,5 +1,5 @@
--- TODO
--- See evals/ba in this repo, as well as https://github.com/tomsmeding/ADBench/blob/101508da93cd4fc80c0ea7a690c6de824c815ca6/tools/Accelerate/src/BA.hs and https://tomsmeding.com/f/master.pdf
+-- TODO: the eval is not yet implemented for horde-ad (just as the ht eval), because it requires quite a bit of work that is not related to AD
+-- For hints, see evals/ba in this repo, as well as https://github.com/tomsmeding/ADBench/blob/101508da93cd4fc80c0ea7a690c6de824c815ca6/tools/Accelerate/src/BA.hs and https://tomsmeding.com/f/master.pdf
 
 {-# LANGUAGE OverloadedLists #-}
 {-# OPTIONS_GHC -fplugin GHC.TypeLits.Normalise #-}
--- a/tools/horde-ad/src/GradBench/Det.hs
+++ b/tools/horde-ad/src/GradBench/Det.hs
@@ -1,10 +1,25 @@
 {-# LANGUAGE OverloadedLists #-}
--- | This is an implementation that directly implements the recursive
--- specification and so is fundamentally so inefficient (something like O(n!)).
+-- | This code directly implements the recursive specification
+-- and so is fundamentally inefficient (something like O(n!)).
 -- Fortunately, the workloads are necessarily tiny so this doesn't OOM.
--- Due to the tiny workloads and, consequently, numerous but tiny tensors,
--- the gradient but especially the primal are much slower than
--- when implemented with lists for haskell-ad.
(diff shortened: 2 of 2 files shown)


Mikolaj:
@samestep and @athas, thank you again for maintaining the benchmark, which greatly helps me understanding horde-ad performance via comparison with other tools. May take some of your time? I'm trying to understand the benchmark results in general and I have some questions.

1. What is the time shown in the cells of the table? Is it the sum over all data sizes of a single runtime of both primal and derivative run? Or not a single runtime, but sum or average of several runs?

2. Is pythorch in lse really managing to have a better asymptotic performance than the manual implementation (and a lot of other tools that have similar results) and this even in the *primal* benchmark? And even if it's not asymptotically better, how on Earth? I'm worried, because maybe it's something like `-ffast-math` and miraculously it works, but other tools universally avoid it for a good reason. Or maybe some tools use more than one core even in the sequential mode, in some clever way? Or is it just the effect of individually handling every combination of the AVX and whatever other CPU extensions (which would be legit, but are they really so powerful?)?

image

3. What's with tensorflow performance in the lse? Is it even better than pythorch asymptotically? Then how come it has 10 times worse score in the benchmark. Shall we let the lines cross by permitting 10 times larger data? What's up with the smallest dataset? Does it compile the kernel or cache things or gather branch prediction data for better JIT? If so, should this time be ignored, by assigning it to "preprocess", as (I'm guessing) in https://github.com/gradbench/gradbench/actions/runs/24759725704/job/72469421996? Or are such long warm-up times really a deficiency of an AD tool and so they should be factored in? Or, on the extreme, should each of the thousands of runs of a single tool be from a cold start?

4. Would it make sense to create an eval even simpler than lse to get to the point where it's obvious no smarts can help at all and so we really measure the bare metal performance of the vector operations in the tools, nothing else? E.g., ensure there's not branching. Maybe just make it one addition? That would help me personally a lot, because I'm suspecting horde-ad is 5 times slower than the rest of the tools on the most trivial vector operations, despite the operations being coded in horde-ad in C using CPU vector instructions. If there's a good reason for the slowness, I'd just take the 5x constant factor into account, but if it's not a constant, it totally obscures any performance data I'm trying to analyze here.

More to come. :)

athas:
> What is the time shown in the cells of the table?

I think it is [the total runtime in seconds](https://github.com/gradbench/gradbench/blob/e03ffb285e177e58784bea39e067c0b90ab5103f/crates/gradbench/src/stats.rs#L142-L199)

> And even if it's not asymptotically better, how on Earth? 

It's probably because it uses various linalg intrinsics. I think the scaling is the same.

> What's with tensorflow performance in the lse?

The tensorflow code is inherited from ADBench and it is *terrible*. What you are seeing is a very high fixed overhead that dominates the runtime.

> Would it make sense to create an eval even simpler than lse to get to the point where it's obvious no smarts can help at all and so we really measure the bare metal performance of the vector operations in the tools, nothing else?

I think `lse` is that eval, and it is hard to imagine any AD problem that doesn't benefit from being clever about hardware. If anything, I'd recommend just ignoring the pytorch/tensorflow tools when you want to understand why the performance is the way it is, because they are operationally very complicated due to the many layers they consist of. Tools like Enzyme and Futhark handle `lse` in a completely straightforward way.

Mikolaj:
> > What is the time shown in the cells of the table?
> 
> I think it is [the total runtime in seconds](https://github.com/gradbench/gradbench/blob/e03ffb285e177e58784bea39e067c0b90ab5103f/crates/gradbench/src/stats.rs#L142-L199)

Yes, I think you are right, it's the time for primal plus the derivative, averaged over all results for the same data sample, summed for all the data samples. The same thing seems to be used as a score and for the display: https://github.com/gradbench/gradbench/blob/e03ffb285e177e58784bea39e067c0b90ab5103f/js/website/src/components/SummaryViz.tsx#L24

Thank you.

PR #668, opened 2026-04-26
Add ffast-math and similar options to horde-ad compilation and also perform other tweaks to horde-ad and haskell-ad

Mikolaj:
The `-ffast-math` option speeds up lse with horde-ad by double-digit percent.  The other options don't seem to do much, but in principle, they may. Is using such options legal? If so and CI shows it doesn't break correctness for any of the benchmarks, I'd set it for horde-ad. Would it make sense to write in some documentation that  `-ffast-math` is fair game or is it too obvious for the AD community (as opposed to the rest of the world)?

I will also perform various other hacks and tweaks to horde-ad and haskell-ad in this PR.

### DIFF
--- a/tools/haskell-ad/gradbench.cabal
+++ b/tools/haskell-ad/gradbench.cabal
@@ -27,4 +27,5 @@ executable gradbench
     hs-source-dirs:   src
     default-language: GHC2021
     default-extensions: OverloadedStrings
-    ghc-options:      -rtsopts "-with-rtsopts=-A500m"
+    -- Idle GC makes benchmarks less deterministic, hence disabled.
+    ghc-options:      -rtsopts "-with-rtsopts=-A500m -I0"
--- a/tools/haskell-ad/gradbench.cabal
+++ b/tools/haskell-ad/gradbench.cabal
@@ -28,4 +28,5 @@ executable gradbench
     default-language: GHC2021
     default-extensions: OverloadedStrings
     -- Idle GC makes benchmarks less deterministic, hence disabled.
-    ghc-options:      -rtsopts "-with-rtsopts=-A500m -I0"
+    -- Also make sure to crash after RAM exhausted instead of swapping, etc.
+    ghc-options:      -rtsopts "-with-rtsopts=-A500m -I0 -M30G"
--- a/tools/haskell-ad/src/GradBench/Det.hs
+++ b/tools/haskell-ad/src/GradBench/Det.hs
@@ -19,8 +19,8 @@ import Data.List qualified as L
 import Numeric.AD.Double qualified as D
 
 data Input = Input
-  { _inputA :: [Double],
-    _inputEll :: Int
+  { _inputA :: ![Double],
+    _inputEll :: !Int
   }
 
 type PrimalOutput = Double
--- a/tools/haskell-ad/src/GradBench/GMM.hs
+++ b/tools/haskell-ad/src/GradBench/GMM.hs
@@ -17,9 +17,9 @@ vMinus :: (Floating a) => V.Vector a -> V.Vector a -> V.Vector a
 vMinus = V.zipWith (-)
 
 data Matrix a = Matrix
-  { matRows :: Int,
-    matCols :: Int,
-    matVals :: V.Vector a
+  { matRows :: !Int,
+    matCols :: !Int,
+    matVals :: !V.Vector a
   }
 
 matRow :: Matrix a -> Int -> V.Vector a
--- a/tools/haskell-ad/src/GradBench/KMeans.hs
+++ b/tools/haskell-ad/src/GradBench/KMeans.hs
@@ -23,9 +23,9 @@ getPoints d v =
   map (getPoint d v) [0 .. (V.length v `div` d - 1)]
 
 data Input = Input
-  { inputD         :: Int,
-    inputPoints    :: V.Vector Double,
-    inputCentroids :: V.Vector Double
+  { inputD         :: !Int,
+    inputPoints    :: !(V.Vector Double),
+    inputCentroids :: !(V.Vector Double)
   }
 
 instance JSON.FromJSON Input where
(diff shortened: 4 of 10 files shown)


athas:
I dislike `-ffast-math` personally because it does too many unstructed things, but I am fine with most of the more precise options that it actually encapsulates (like `-fno-math-errno`).

GradBench doesn't have a ban against using `-ffast-math` as long as things reliably validate, and it would be weird to have such a ban, as it raises questions about how to enforce it against non-C languages.

Mikolaj:
> I dislike `-ffast-math` personally because it does too many unstructed things, but I am fine with most of the more precise options that it actually encapsulates (like `-fno-math-errno`).

Right. Which may be one of the reasons ox-arrays doesn't set it (nor horde-ad that depends on it). Therefore, I've only enabled it for gradbench, in the `cabal.project` file.

> GradBench doesn't have a ban against using `-ffast-math` as long as things reliably validate, and it would be weird to have such a ban, as it raises questions about how to enforce it against non-C languages.

Shall I mention this somewhere to level up the playing field? I'm seeing "clang: warning: argument '-Ofast' is deprecated; use '-O3 -ffast-math' for the same behavior" in an old PR, so maybe OCaml already does this and maybe others do, too, though I can't find and explicit `-Ofast' option anywhere.

I've found that `-ffast-math` speeds up `lse` and `kmeans` by around 2 times for horde-ad, so that's already a big portion of the 5 times slowness I set out to explain in https://github.com/gradbench/gradbench/pull/667. With other benchmarks, it's swallowed by noise. The lse comparison is https://github.com/gradbench/gradbench/actions/runs/24957454605/job/73079668058?pr=668 vs https://github.com/gradbench/gradbench/actions/runs/24921971554/job/72997371938 and kmeans is https://github.com/gradbench/gradbench/actions/runs/24957454605/job/73079668026?pr=668 vs https://github.com/gradbench/gradbench/actions/runs/24921971554/job/72997371905.

Probably some of the noise is unavoidable, coming from the performance variability of the cloud, isn't it? But I fear a portion may come from GC. Randomly, a major GC may trigger (e.g., due to some delay and the idle GC mechanism) in test 44 of a benchmark or in test 46 or it may not be required until the benchmark is done and then no GC is ever needed. Consequently, in one run 44 seems expensive but 46 cheap, in another it's the reverse and in yet another where GC almost triggers but not quite, the whole tool randomly gets a better score. What would be the fair way to solve this? I guess, starting the program afresh for each data size? If so, to avoid slowing down CI by 50*10 startup times, we could simulate this by forcing a major GC before each each data size testing starts (before, not after, to avoid the last GC that would be the most expensive but also not needed at all). This looks fishy, because these GCs are not included in the timings, but it's the same behaviour that tools (not libraries) started afresh for each data size do (are there any such tools?). If that's a good idea, it's a one-liner, so I could implement it in this PR for horde-ad and haskell-ad.

Mikolaj:
Precedents for GC before the clock is started are below, though this is delicate and what is proper may depend on whether the intended atomic benchmark is the whole run of, say, 'lse', or separately each of the runs 2 to 40 of `lse`. That's because running GC inside a benchmark and not counting the time spent on it is obviously cheating.

https://github.com/haskell/criterion/blob/42a7595012306f899bfc2039a5db114757c3dcef/criterion-measurement/src/Criterion/Measurement.hs#L292

https://github.com/Bodigrim/tasty-bench/blob/659e800cdaa52e9b539a368f60a96149e66449a4/src/Test/Tasty/Bench.hs#L1175

Mikolaj:
Oh, wow, it turns out that stopping idle GC (`-I0`) impacts the `lse` score of horde-ad significantly and forcing the input makes the `lse` score of horde-ad 3 times higher in my local test. OTOH, haskell-ad is unaffected, unless the noise hides the effect, given that the absolute times are orders of magnitude higher. So, apparently, creating, encoding, parsing and allocating the data is 2 times more expensive than the real `lse` computation in horde-ad. Given how expensive JSON is, that's quite plausible. Hmm, and the heap inflated in this expensive processing of samples is sometimes GCd while the measurement clock is running.

Mikolaj:
CI almost passed after one restart, so let me experiment with the forced GC idea now. In my local `lse` test it speeds up horde-ad just a little and haskell-ad some more, which is expected, since horde-ad barely allocates in `lse` (mostly when processing JSON, probably).

@athas: shall I also add to this PR the following hack: https://github.com/gradbench/gradbench/pull/665#issuecomment-4320645109  ?

athas:
Forcing the input is completely uncontroversial and should be done (and not measured).

I personally have little problem with forcing (and not measuring) a major GC before/after every workload. That seems reasonable enough, as long as it is used in the service of having more realistic numbers.

> I've found that -ffast-math speeds up lse and kmeans by around 2 times for horde-ad

I *suspect* (because I have observed it before) that this is because `-ffast-math` allows for open-coded or instruction based code generation for certain mathematical functions, in the case of `kmeans` probably square roots, instead of making a much slower call to `libm.` That is, the performance increase is not due to the dangerous parts of `-ffast-math`, but things like not setting errno. Anyway, the cause doesn't matter, if it validates it's good.

athas:
And feel free to add whatever hacks and RTS tunings you feel appropriate. All the C implementations are full of similar things for preallocating memory etc. Do make a note if performance crucially depends on GHC RTS arcana.

Mikolaj:
OK, I've now pushed it all and I also documented each tweak near its code. For the heap limit, I've started conservatively at 30G, but we'd ideally set the real RAM available after OS overhead. Any idea how much RAM the machines have?

Mikolaj:
> Any idea how much RAM the machines have?

I have trouble finding that information. I guess 16GB is most likely, so 12G or 14G may be a better limit. I will switch to it next time I have to touch a commit in this PR in order to restart CI.

Mikolaj:
Actually, with 30G heap limit, kmeans for haskell-ad ended with "succeeded 8 minutes ago in 4m 44s", which is ideal, so maybe 30G is fine, because with the copying GC half of the heap is empty. So the limit probably triggers the moment the heap grows to 30G, while the live set is still around 15G and so (barely) fits in RAM.

So this PR is ready for review.

Mikolaj:
And even CI is green. :)

athas:
@samestep Could we add @Mikolaj as a contributor so his PRs will get graphs?

samestep:
Absolutely!

@Mikolaj I sent you an invite to join @gradbench/team. (I also accidentally sent you an invite to join this repo directly, but then canceled that after I remembered about the organization-level team we've been using.)

Mikolaj:
Thank you, I'm honoured.

PR #669, opened 2026-04-29
Lower the heap limit for the Haskell tools to prevent hanging containers and also tweak Haskell and OCaml things for CI

Mikolaj:
Apparently, the 30G limit sometimes causes a gentle quick OOM but sometimes it hangs hard: https://github.com/gradbench/gradbench/actions/runs/25091053615/job/73544366949, so I'm bisecting until I find a limit where all tests pass but the haskell-ad/kmeans  OOM is quick.

### DIFF
--- a/tools/haskell-ad/gradbench.cabal
+++ b/tools/haskell-ad/gradbench.cabal
@@ -29,4 +29,4 @@ executable gradbench
     default-extensions: OverloadedStrings
     -- Idle GC makes benchmarks less deterministic, hence disabled.
     -- Also make sure to crash after RAM exhausted instead of swapping, etc.
-    ghc-options:      -rtsopts "-with-rtsopts=-A500m -I0 -M30G"
+    ghc-options:      -rtsopts "-with-rtsopts=-A500m -I0 -M14G"
--- a/tools/haskell-ad/src/Main.hs
+++ b/tools/haskell-ad/src/Main.hs
@@ -161,10 +161,18 @@ main = forever loop
             BS.putStrLn . BS.toStrict . JSON.encode . JSON.object $
               ("id", toJSON (msgId msg)) : vs
             hFlush stdout
+          onRunError :: SomeException -> IO a
+          onRunError e = do
+            reply
+              [ ("success", toJSON False),
+                ("error", toJSON $ "tool exception:\n" <> take 1000 (show e))
+              ]
+            throw e
       case msg of
         MsgStart _ _ -> reply [("tool", "haskell-ad")]
         MsgDefine _ mod -> reply [("success", toJSON $ knownModule mod)]
         MsgEvaluate _ mod fun input ->
+          (`catch` onRunError) $
           case L.lookup (mod, fun) modules of
             Nothing ->
               reply
--- a/tools/horde-ad/cabal.project
+++ b/tools/horde-ad/cabal.project
@@ -5,7 +5,9 @@ packages: gradbench.cabal
 -- The other options sound like plausible speedups looking at their documentation
 -- and apparently don't lower performance at least.
 package ox-arrays
-  flags: +nonportable-simd
+  -- CI seems to sometimes compile on a different CPU than the binary is run on
+  -- so this optimisation has to be disabled
+  -- flags: +nonportable-simd
   ghc-options: -optc-ffast-math -optc-fassociative-math -optc-fno-signed-zeros -optc-fno-trapping-math -optc-freciprocal-math
 
 ghc-options: -fexcess-precision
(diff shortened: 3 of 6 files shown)


Mikolaj:
15G still causes haskell-ad/kmeans to take 15min and possibly more, so let me try 12G.

Mikolaj:
12G is fine for haskell-ad and kmeans, but not enough for horde-ad and det. Let's see if 14G would be just right for both or if we need the two to diverge.

Mikolaj:
It worked, so maybe 14G max heap is the sweet spot, we'll see over time. 

Ready for review.

Mikolaj:
BTW, there are possibly genuine horde-ad gradient-generating failures, unrelated to heap size and RTS options, at 

https://github.com/gradbench/gradbench/actions/runs/25147619835/job/73733267599#step:5:95

and

https://github.com/gradbench/gradbench/actions/runs/25147619835/job/73733267929#step:5:28

from today's nightly (still running). I will try to add better diagnostics so that next time we can see what's failing. If this here PR is merged (pretty-please), I will add this in a next PR, otherwise, I'd sneak it here.

Mikolaj:
OK, the extra exception handling added and it apparently doesn't break CI, though so far no Haskell exceptions appear for us to verify that error reporting improves. @athas: please kindly review.

Mikolaj:
Since CI currently crashes for OCaml, swallowing errors, let me try the same trick for OCaml.

Mikolaj:
Too bad, it didn't help for OCaml, we are still getting only

```
running eval kmeans
   with tool ocaml
Broken pipe (os error 32)
 outcome error
expected undefined
Error: Process completed with exit code 1.
```

so either this is an OS error that is not turned into an OCaml exception (my OCaml is very rusty, I don't know if that's likely; I guess it can happen if the binary is somehow wrong, compiled on a different architecture (or with different CPU extensions?)  than it runs on) or it happens outside of the main computation (`f input`). If the latter, maybe we could wrap *everything* in a `try...with` bracket. Or it may be an error in the test harness, not the main OCaml executable at all. No idea.

It may still make sense to merge the extra error reporting for OCaml and monitor in CI if it ever causes problems (so far, it doesn't seem so --- some OCaml tests passed fine).

So, again, the PR is ready for review.

athas:
I'm impressed at your willingness to debug through GA latency (and unpredictability). It is by far the least comfortable "deployment" environment I know of.

Mikolaj:
I happen not to have an OCaml compiler installed, so CI is my only debugger. :)

samestep:
@Mikolaj may I recommend getting OCaml via the Nix shell provided by this repo? I'm guessing you'll find the reduced latency worthwhile. (But no pressure if you'd prefer to use CI!)

athas:
FWIW, I think the changes in this PR are good, or at least harmless. I have a hard time forming any real emotion about playing with GHC RTS parameters. I will merge once/if CI succeeds (or if it only fails for unrelated reasons).

samestep:
@athas I'll defer to your judgment since Haskell and OCaml are not my areas of expertise.

Mikolaj:
> @Mikolaj may I recommend getting OCaml via the Nix shell provided by this repo? I'm guessing you'll find the reduced latency worthwhile. (But no pressure if you'd prefer to use CI!)

Yes, thank you, that sounds rational. I guess, in addition to the thrill of doing something irrational, I'm not installing OCaml to help my future self remember that I'm not responsible for the OCaml tool, despite the token effort I invested in it. I tend to pile on related responsibilities and lose track of the original goal (which was not the Haskell portion of gradbench nor even the horde-ad library to start with). So, this is kind of a life hack. :)

Mikolaj:
I'm happy to report that the hacks probably helped for haskell-ad and horde-ad (to be seen, but so far so good), and sad to report than none of my two hacks even offered any new debug info about the OCaml failures. If my last horde-ad fix was really what fixed it (and not just the CI randomness), then our CI has the problem that `-march=native` is an invalid optimization, because apparently (if this is all not just a fluke) the binaries are sometimes built on different CPUs than what they are run on. I wonder if that's also what kills the OCaml processes, though the symptoms are different --- horde-ad was crashing once it started using fancy CPU vector instructions and OCaml systematically crashes outright and that even for evals that it does not implement. Let's see how the subsequent CIs are going to behave and if the patterns persist.

Let me restart CI to have one more data point (but feel free to merge already --- the last CI does not contain anything suggesting this PR is wrong).

athas:
`-march=native` has proven problematic in the past, for sure.

Mikolaj:
> `-march=native` has proven problematic in the past, for sure.

But you seem to get away with it?

```
tools/futhark/run.py:CFLAGS = "-O3 -march=native -fno-math-errno"
```

Oh, but that's only for `message["kind"] == "define"` and maybe there's a guarantee that then `message["kind"] == "evaluate"` is run on the same CPU. Unfortunately, horde-ad can't afford to repeat most of the compilation for each eval --- unless there's a cache that stores the compiled main library, one copy for each hash of CPU capabilities already used.

Anyway, CI looks fine, so let me merge before the nightly starts. Thank you for the reviews and hints!

PR #671, opened 2026-05-01
Give horde-ad a weaker but safer variant of -march=native

Mikolaj:
Unfortunately, the removal of `-march=native` was not the cause of the 2-times slowdown of a few horde-ad evals, because the slow and the fast result seem to appear at random, even for different evals within the same CI run and regardless of the compiler flags I set. 

I still think it must be caused by different CPU kinds used by CI for different evals and it may or may not be additionally affected by what CPU the binary was built and optimized for originally (because for Haskell, we build the binary once and use it for all evals; it's hard to do otherwise without repeating the whole long compilation for each eval).

Regardless, the tweak in this PR seems safe and it potentially could improve the performance, even though it probably doesn't or at least there's too much noise to tell.

Please kindly review.

### DIFF
--- a/tools/horde-ad/cabal.project
+++ b/tools/horde-ad/cabal.project
@@ -2,12 +2,20 @@ packages: gradbench.cabal
 
 -- Of all the options below, nonportable-simd helps across the board,
 -- while ffast-math helps immensely in lse and kmeans, but much less elsewhere.
--- The other options sound like plausible speedups looking at their documentation
--- and apparently don't lower performance at least.
+-- The other options sound like plausible speedups looking at their
+-- documentation and apparently don't lower performance at least.
+
+-- CI seems to sometimes compile on a different CPU than the binary is run on
+-- so this optimisation that contains "-march=native" has to be disabled:
+-- package ox-arrays
+--   flags: +nonportable-simd
+-- but this weaker optimization seems safe enough for github CI:
+if arch(x86_64)
+  package ox-arrays
+    ghc-options: -optc-march=x86-64-v3
+
 package ox-arrays
-  -- CI seems to sometimes compile on a different CPU than the binary is run on
-  -- so this optimisation has to be disabled
-  -- flags: +nonportable-simd
+  -- These seem safe always:
   ghc-options: -optc-ffast-math -optc-fassociative-math -optc-fno-signed-zeros -optc-fno-trapping-math -optc-freciprocal-math
 
 ghc-options: -fexcess-precision


Mikolaj:
BTW, @samestep, is it possible to see a nice visualization of stats for a PR like that?

athas:
Open the old and new graphs in different browser tabs and flip back and forth quickly.

The reason there is no graph in this PR, despite all tests succeeding, is AFAIK because the source branch is from a different repository.

athas (approved):

Mikolaj:
> The reason there is no graph in this PR, despite all tests succeeding, is AFAIK because the source branch is from a different repository.

Aha! Thank you.

Mikolaj:
BTW, if you don't mind, I plan to restart failing jobs in the next few nightlies in order to find one where CPUs align and I can compare the best case of horde-ad performance against the other tools. The failed jobs should be retrievable without much effort if they turn out to be needed for debugging (I'm guessing most of the failures these days are due to transient network or server failures and also due to OCaml, probably akin to the `-march=native` problem I fixed in horde-ad).

athas:
Sure, knock yourself out. But it might be easier to run it locally. It's *much* more reliable than spinning the GitHub actions lottery wheel, and you'll be certain it's all on the same machine. This is how you run every single eval on every single tool:

```
$ ./gradbench repo run
```

And if you just want to run `horde-ad`:

```
$ ./gradbench repo run --tool horde-ad
```

You'll probably also want to use the `--output` option to get machine-readable data. It's all documented fairly well.

Mikolaj:
Thank you, that's not as scary as I thought and so I may do just that. However, in addition, I'd like to brag about horde-ad stats and there linking to your shiny public page is the best way.

Mikolaj:
Some initial feedback: 

> ```
> $ ./gradbench repo run
> ```

This works great (specifically, `./gradbench repo run -o out --timeout 600`), but somewhere during the run, two julia processes go rogue, use two full cores, take ~30G of RAM and stay there even after the whole run is finished:

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                                                                
 142983 root      20   0 1693348 470216 149940 R 100.0   0.7  40:33.33 julia                                                                                                                                                                  
 149590 root      20   0   34.3g  33.1g 152908 R 100.0  52.7  15:09.48 julia                                                                                                                                                                  
```

I wonder if this happens in CI as well. It surely skews the results of the tools that run afterwards when invoked locally.

athas:
That's interesting, I didn't even know that was possible, since the Docker containers are supposed to be shut down.

Mikolaj:
I think this is not a fluke, because it happened to me 2/2, once in a run without `--timeout`, which I had to abandon via C-c due to tools filling all of my available RAM repeatedly, and again, after a machine reset, with `--timeout`, which ran to the end ("running eval saddle with tool zygote") and exited without any error message.

athas:
Do you know which of the Julia tools do this? Maybe `ps aux` will reveal the culprit.

Mikolaj:
Too late, I killed the rogue processes and didn't think to show their commandline.

Mikolaj:
Let me start anew and in a few hours I will tell you the commandline of the remaining julia processes.

athas:
Thanks! I tried yesterday but my machine crashed, so you are probably right that something is awry. I think it is a recent change, which is of course a disturbing thing to consider, but not unexpected given that our Docker images are not fully reproducible.

Mikolaj:
This is still running (currently at kmeans) but there's already one stuck process:

```
~/r/gradbench$ ps -aux | grep julia
root      206731 99.6 19.5 22184580 12885144 ?   Rsl  15:18  15:33 julia --project=. run.jl
```

Can I somehow query it more precisely?

Mikolaj:
`ps -aux --forest` does not seem very helpful:

```
root      206675  0.0  0.0 1233616 10836 ?       Sl   15:18   0:00 /usr/bin/containerd-shim-runc-v2 -namespace moby -id b244e3ef2b8b961551c3c093331281287107fe8665074289bac047b9d91dae95 -address /run/containerd/containerd.sock
root      206731 99.7 26.1 22533516 17213384 ?   Rsl  15:18  18:06  \_ julia --project=. run.jl
```

athas:
Yeah, that's useless. I guess that container ID is just some meaningless hash as well.

Mikolaj:
yes, it seems so:

```
~$ docker container ls
CONTAINER ID   IMAGE                                  COMMAND                  CREATED          STATUS          PORTS     NAMES
14fbc8f7c5b5   ghcr.io/gradbench/tool-ad-hpp:latest   "python3 python/grad…"   31 seconds ago   Up 30 seconds             condescending_leavitt
ba90490f5728   ghcr.io/gradbench/eval-lse:latest      "uv run --no-sync py…"   31 seconds ago   Up 30 seconds             musing_banach
```

```
~$ docker container inspect 547a235c35ca65f8a082fa4ffb2e7c7664f4f77b7953457d4b1332b1edd3b031
[]
Error response from daemon: No such container: 547a235c35ca65f8a082fa4ffb2e7c7664f4f77b7953457d4b1332b1edd3b031
``

**Comment by Mikolaj** [comment]
Hah, the process finished a moment ago. So maybe each of these does finish eventually, it just takes a long time and overlaps with other benchmarks. Probably some more julia processes would be spawned before the full gradbench run is done and some will live longer than the main gradbench process, but apparently not the one I posted above. It may be somewhat random, too.

**Comment by athas** [comment]
That is possible: when something times out, we may actually leave the tool process running, because we do not have a way of killing it. (Or just didn't implement that part.)

**Comment by Mikolaj** [comment]
I caught forwarddiff-jl red-handed. It just timed out at eval ode and created

```
root      246766  0.0  0.0 1233872 10844 ?       Sl   16:44   0:00 /usr/bin/containerd-shim-runc-v2 -namespace moby -id bfd15b48fd1bbb166dd375ba6e5d3fb0b54f2d476a5972c652ce6e94963429c2 -address /run/containerd/containerd.sock
root      246822 99.8  0.8 1763308 555324 ?      Rsl  16:44  22:05  \_ julia --project=. run.jl
```

that persisted while futhark and then haskell-ad were running ode.

**Comment by Mikolaj** [comment]
And now `reversediff-jl` did the same with ode, grandhbench finished, the two julia processes remain still.

**Comment by Mikolaj** [comment]
And one of them could be killed with `sudo kill ###`, the other required `kill -9`.

**Comment by athas** [comment]
`reversediff-jl` is one of the ones that time out frequently. This suggests we need to look into how we shut town tools after timeouts.

**Comment by Mikolaj** [comment]
> `reversediff-jl` is one of the ones that time out frequently. This suggests we need to look into how we shut town tools after timeouts.

Yes, please! Let me know if I can test any attempted solutions.

**Comment by athas** [comment]
I can reproduce locally, and the easiest way to reproduce is this:

```
$ ./gradbench repo run --eval det --tool reversediff-jl --timeout 10
```

On my machine this leaves a Julia process around for about ten minutes.

PR #672, opened 2026-05-06
Horde-ad experiments

Mikolaj:


Mikolaj:
kmeans for haskell-ad sometimes ends with `error` and rarely with `timeout` (the latter, e.g., in https://github.com/gradbench/gradbench/actions/runs/25510158301/job/75082016601#step:5:46). Is there an easy way to permit both outcomes? I suppose, at worst, I could add a timer thread to haskell-ad that errors out after the given time, but that's a hack and it hides some information about a test run in question.

athas:
Yes, you can put `error+timeout` as the expected outcome.

Mikolaj:
FYI, I wasn't able to fix the two last Nightly runs that failed due to ocaml by restarting only failed eval jobs several times. Probably the ocaml jobs got mis-compiled, e.g., maybe compiled with `-march=native` on an exotic CPU that rarely appears in CI runs, and then restarting eval jobs apparently doesn't help.

Mikolaj:
But, as far as I can see from a quick WWW search, dune, the OCaml build tool, doesn't set `-march=native`. So it may be a different kind of mis-compilation.

athas:
I have also noticed this. The OCaml tool artifact seems to often end up corrupted somehow.

Mikolaj:
I'm trying to prevent CI from breaking whole runs due to ocaml miscompilations, but for some reason `success+error` fails --- it says "tools/ocaml/evals.txt": invalid outcome for eval "gmm": invalid outcome "success": Matching variant not found".

```
diff --git a/tools/ocaml/evals.txt b/tools/ocaml/evals.txt
index 4263da4b..6ff7252e 100644
--- a/tools/ocaml/evals.txt
+++ b/tools/ocaml/evals.txt
@@ -1,3 +1,12 @@
-gmm
-hello
-lse
+ba undefined+error
+det undefined+error
+gmm success+error
+hello success+error
+ht undefined+error
+kmeans undefined+error
+llsq undefined+error
+lse success+error
+lstm undefined+error
+ode undefined+error
+particle undefined+error
+saddle undefined+error
```

athas:
Yes, for technical reasons, you cannot mix `success` with an error state. This is because the model is success *or* a set of acceptable errors.

Mikolaj:
Thank you for the explanation. So, I tried to let CI ignore ocaml miscompilations at least in the cases where `undefined+error` suffices, e.g., `particle undefined+error`, but it consistently fails with `no results for tool ocaml` for particle (strangely, others are fine), as in https://github.com/gradbench/gradbench/actions/runs/25821255311/job/75983024488.

Mikolaj:
And this time the result was `undefined`, so probably no mis-compilation occurred, so it's not a result of an OCaml corruption, most probably: https://github.com/gradbench/gradbench/actions/runs/25821255311/job/75982414165

Mikolaj:
Shall I report this as a bug in a separate issue?

athas:
Feel free to report the OCaml flakiness as an issue. I really have no idea what is happening.

Mikolaj:
> Feel free to report the OCaml flakiness as an issue. I really have no idea what is happening.

But I meant the "no results for tool ocaml" message and failure, which is shown for an eval that is not defined for ocaml: https://github.com/gradbench/gradbench/pull/672#issuecomment-4451050372

This makes even my attempted half-workaround fail.

Regardless, on your advice, I will report the ocaml flakiness.

Mikolaj:
I've now tested locally a couple of variants of `-optc-march` and `-optc-mtune` for horde-ad and I can't see much of a difference. However, in CI, when the compilation CPU and runtime CPU differ, the difference may manifest, so let's let the tweaks in the PR stay. The commits in this PR that are not about performance should be helpful regardless of CPUs used. Please kindly review and merge.

PR #674, opened 2026-07-22
Fix Nix shell

samestep:
It got broken because we build Floretta from source and crates.io started giving 403s in some cases, so we bump to a newer Nixpkgs that knows how to ask crates.io more politely.

This also updates Nixfmt and Clippy, so I did a `./gradbench repo lint --fix` to fix those.

### DIFF
--- a/crates/gradbench/src/util.rs
+++ b/crates/gradbench/src/util.rs
@@ -71,7 +71,7 @@ pub fn nanostring(nanoseconds: u128) -> String {
     }
 }
 
-pub fn lock<T>(mutex: &Mutex<T>) -> MutexGuard<T> {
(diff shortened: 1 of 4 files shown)


athas:
This works for me, and CI does not use `nix-shell` meaning those errors are just noise, so I am going to merge this PR.

Also, we have a somewhat ridiculous shell.nix that pulls in various esoterica. You need this to *run* the tools outside of Docker, but it is excessive if you just want enough to work on `gradbench` itself. Perhaps we should factor it better.

samestep:
I don't mind pulling everything into the dev shell, but I wouldn't be opposed to factoring it to also have leaner shell options for people who'd prefer those.

